PR #4, opened 2018-12-05
Changing method call for cube cutting

arfon:
:wave: @ceb8 - the docs as written didn't work for me here. I think we need to call `cube_cut`.


ceb8:
Thanks @arfon!

PR #6, opened 2018-12-07
Bugfix: the aperture extension must use integers

barentsen:
Thank you for providing the fabulous TESScut service.  It is impressively fast and user-friendly!

I noticed that the files returned by TESScut contain an aperture mask image extension (HDU 2) which contains floating point values.  In contrast, the standard TESS/Kepler target pixel files contain integers in this extension.

This is a minor issue because this extension represents a bitmask image, and therefore *lightkurve* (and likely other data analysis tools) expect to encounter integers here. I believe this PR fixes the issue!


ceb8:
Thanks @barentsen!

PR #7, opened 2018-12-07
Set the `CREATOR` keyword to `astrocut`

barentsen:
Currently the TESScut service returns Target Pixel Files with a header which suggests that they were created using the FFI exporter of the official pipeline, cf.:
```
CREATOR = '113 FfiExporter'    / pipeline job and program used to produce this f
PROCVER = 'spoc-3.3.31-20180902' / SW version 
```

This PR proposes to set the `CREATOR` keyword to contain `astrocut`, which enables TPFs generated using astrocut/tesscut to be identified more easily.

Again, thank you for providing the fantastic TESScut service!

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -14,6 +14,7 @@ from time import time
 import os
 import warnings
 
+from . import __version__
 from .exceptions import InputWarning, TypeWarning, InvalidQueryError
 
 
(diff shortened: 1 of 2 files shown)


astrocut/cube_cut.py
```diff
@@ -318,6 +318,9 @@ def _update_primary_header(self, primary_header):
         """
 
         # Adding cutout specific headers
+        primary_header['CREATOR'] = ('astrocut', 'software used to produce this file')
+        primary_header['PROCVER'] = (None, 'software version')
```
ceb8: Can you make this actually fill in Astrocut's version number?

barentsen:
I changed this PR to actually populate the header with the astrocut version.  Most Travis tests pass, though one fails and I don't think I understand why!

ceb8:
Thanks @barentsen!

PR #10, opened 2019-04-05
Uniform formatting on target pixel file names

nksaunders:
In `cube_cut`, the string formatting for target coordinates in the downloaded filename was not specified. The minor change in this PR ensure that TESScut filenames are saved with the coordinates in the form of floats with consistent length. This makes it easier to check locally if a given target has already been downloaded.


ceb8:
Thanks @nksaunders!

PR #12, opened 2019-05-10
WCS bugfix

ceb8:
Sometimes the TESS pipeline fails to produce a WCS object, in this case the FFI lacks the WCS header keywords.  This is a problem for Astrocut in two situations:

1. The FFI used to set up the header keywords table in the cube file is missing the WCS header keywords, in which case the cube as a whole is missing those keywords.
2. The FFI used to make the WCS information for the cutout lacks the WCS keywords.  In this case a WCS object cannot be created so the cutout cannot proceed.

This PR solve both of these problems with the following changes:

- In the make_cube file the WCS keyword "CTYPE1" is checked and must be present for an FFI to be used as the template for making the header table.
- In the cube_cut file the WCS keyword "CTYPE1" is checked and must be present  for the row of header information to be used to make the cutout WCS.

Additionally I added a catch for the NoResolve exception when finding the cutout location and instead raise a clearer error message informing the user that their cutout is not on the FFI in question.

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -88,13 +88,13 @@ class CutoutFactory():
             One row from the cube image header data table.
         """
 
-        data_ind = int(len(table_data)/2)
+        data_ind = int(len(table_data)/2)  # using the middle file for table info
         table_row = None
 
         # Making sure we have a row with wcs info
         while table_row is None:
             table_row = table_data[data_ind]
-            ra_col = int([x for x in table_header.cards if x[1] == "CTYPE1"][0][0].replace("TTYPE","")) - 1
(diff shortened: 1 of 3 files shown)


astrocut/cube_cut.py
```diff
@@ -88,7 +88,21 @@ def _parse_table_info(self, table_header, table_row):
             One row from the cube image header data table.
         """
 
-        # Turning the table row inot a new header object
+        data_ind = int(len(table_data)/2)  # using the middle file for table info
+        table_row = None
+
+        # Making sure we have a row with wcs info
+        while table_row is None:
```
phicc: May want to add a check in case no table row is found

ceb8: Done.

Issue #26, opened 2019-11-25
Typo in error message


jeffchamblee:
"Cutout contains **to** data!" in cutout.py. Should be "Cutout contains no data." "To data" occurs in four places.

Issue #16, opened 2019-08-09
Minor errata


ceb8:
Issue to capture small non-urgent things that I notice while adding new functionality:

#### cube_cut.py
- `_get_cutout_limits` docstring parameters is wrong (center_coord is no longer a param)
- `target_pixel_file = os.path.join(output_path, tpf_name)` is indented one too far
- Get rid of the overwrite=True or make it an argument in `cube_cut`?

PR #28, opened 2019-12-20
Error handling

ceb8:
Adding error handling so that if one file in a list for fits_cut or image_cut raises an error, the user still gets a cutout with the rest of the files.  A warning is raised showing which files caused a problem and what the error was.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ env:
         - PYTHON_VERSION=3.7
         - NUMPY_VERSION=stable
         - ASTROPY_VERSION=stable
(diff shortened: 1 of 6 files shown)


ceb8:
This PR also addresses some of the smaller issues from https://github.com/spacetelescope/astrocut/issues/16.

ceb8:
This PR also addresses https://github.com/spacetelescope/astrocut/issues/26.

ceb8:
This PR also updates the fits handling to use `with fits.open()` syntax so that fits files will be properly closed if an error is raised while they are in use.

Issue #22, opened 2019-11-15
Update fit_wcs_from_points to match astropy


ceb8:
When `fit_wcs_from_points` went into astropy its call pattern changed, the local version in astrocut needs to be updated to match so that astrocut can use the astropy version if it is available and it's own if it isn't.

Additionally testing needs to be done to make sure the accuracy of the fit wcs remains the same.

PR #29, opened 2020-04-23
Updating wcs fitting to match astropy

ceb8:
The PR resolves issue https://github.com/spacetelescope/astrocut/issues/22.

- Updated the local `fit_wcs_from_points` to match what is in Astropy.
- Added additional residual checking.
- Opened bugfix PR to add the same residual checking in Astropy (https://github.com/astropy/astropy/pull/10155).
- Added check so that for Astropy version >=4.0.1 the Astropy function will be used preferentially (this will be the version my bugfix goes into)

### DIFF
--- a/astrocut/utils/wcs_fitting.py
+++ b/astrocut/utils/wcs_fitting.py
@@ -20,15 +20,28 @@ import numpy as np
 from astropy import units as u
 from astropy.wcs.utils import celestial_frame_to_wcs
 
-def _linear_transformation_fit(params, lon, lat, x, y, w_obj):  # pragma: no cover
-    """ Objective function for fitting linear terms."""
-    pc = params[0:4]
+def _linear_wcs_fit(params, lon, lat, x, y, w_obj):
+    """
+    Objective function for fitting linear terms.
+
+    Parameters
+    ----------
+    params : array
+        6 element array. First 4 elements are PC matrix, last 2 are CRPIX.
+    lon, lat: array
+        Sky coordinates.
+    x, y: array
+        Pixel coordinates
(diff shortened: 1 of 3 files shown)


astrocut/utils/wcs_fitting.py
```diff
@@ -20,197 +20,249 @@
...
+    params : array
+        6 element array. First 4 elements are PC matrix, last 2 are CRPIX.
+    lon, lat: array
+        Sky coordinates.
+    x, y: array
+        Pixel coordinates
+    w_obj: `~astropy.wcs.WCS`
+        WCS object
+        """
+    cd = params[0:4]
```
scfleming, line 38: A little dangerous to access parameters directly on indexes, if params[] changes structure in the future this would need to be updated and could be forgoten about.  Minor comment though.

astrocut/utils/wcs_fitting.py
```diff
@@ -20,197 +20,249 @@
...
-    the keyword argument will override, including the CD/PC convention.
-    NAXIS1 and NAXIS2 will set as the span of the input points, unless an input WCS is
-    provided.
-    All output will be in degrees.
+def _sip_fit(params, lon, lat, u, v, w_obj, order, coeff_names):
+
+    """ Objective function for fitting SIP.
+     Parameters
+    -----------
+    params : array
```
scfleming, line 57: Same comment on passing index of parameter values with an assumed ordering...

ceb8:
Thanks @scfleming! I'm not going to change for this PR, but those are great comments, and I'll make a note to look into submitting an improvement to astropy to future proof in that way.

PR #30, opened 2020-04-28
Streamline WCS fitting

ceb8:
This pertains to TESS cutouts.
Currently when fitting a linear WCS to a cutout we use every single pixel in the fit. This gets inefficient for large cutouts and does not make the fit appreciable better.
This PR limits the number of pixels used to a maximum of 100, spread evenly throughout the cutout.

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -13,7 +13,7 @@ from itertools import product
 
 # Note: Use the astropy function if available
 import astropy
-if astropy.utils.minversion(astropy,"4.0.1"):
+if astropy.utils.minversion(astropy,"4.0.2"):
     from astropy.wcs.utils import fit_wcs_from_points
 else:
     from .utils.wcs_fitting import fit_wcs_from_points
@@ -273,24 +273,32 @@ class CutoutFactory():
 
         # Getting matched pixel, world coordinate pairs
         # We will choose no more than 100 pixels spread evenly throughout the image
+        # Centered on the center pixel.
         # To do this we the appropriate "step size" between pixel coordinates
-        # (i.e. we take every ith pixel in each row/column)
+        # (i.e. we take every ith pixel in each row/column) [TODOOOOOO]
         # For example in a 5x7 cutout with i = 2 we get:
         #
-        # xoxoxox
+        # xxoxoxx
         # ooooooo
-        # xoxoxox
+        # xxoxoxx
         # ooooooo
-        # xoxoxox
+        # xxoxoxx
         #
         # Where x denotes the indexes that will be used in the fit.
-
         y, x = cutout_shape
         i = 1
         while (x/i)*(y/i) > 100:
             i += 1
             
-        pix_inds = np.array(list(product(list(range(0,x,i)), list(range(0,y,i)))))
+        xvals = list(reversed(range(x//2, -1, -i)))[:-1] + list(range(x//2, x, i))
+        if xvals[-1] != x-1: xvals += [x-1]
+        if xvals[0] != 0: xvals = [0] + xvals
+        
+        yvals = list(reversed(range(y//2, -1, -i)))[:-1] + list(range(y//2, y, i))
+        if yvals[-1] != y-1: yvals += [y-1]
+        if yvals[0] != 0: yvals = [0] + yvals
+        
+        pix_inds = np.array(list(product(xvals,yvals)))
         world_pix = SkyCoord(cutout_wcs.all_pix2world(pix_inds, 0), unit='deg')
 
         # Getting the fit WCS
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -15,15 +15,15 @@ from astropy.io import fits
 from astropy.coordinates import SkyCoord
 from astropy import wcs
 
+from . import __version__ 
+from .exceptions import InputWarning, TypeWarning, InvalidQueryError
+
 # Note: Use the astropy function if available
 import astropy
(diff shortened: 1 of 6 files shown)


astrocut/cube_cut.py
```diff
@@ -272,17 +272,26 @@ def _fit_cutout_wcs(self, cutout_wcs, cutout_shape):
         """
 
         # Getting matched pixel, world coordinate pairs
-        pix_inds = np.array(list(product(list(range(cutout_shape[1])), list(range(cutout_shape[0])))))
+        # We don't want more than ~100 pairs, and we want them spread evenly through the iamge
+
+        y, x = cutout_shape
+        i = 1
+        while (x/i)*(y/i) > 100:
```
scfleming, line 279: Suggest adding an in-line comment on what you are doing here.  Took me a while to figure it out (if I got it right), I think this line is determining how large the spacing should be to ensure 100 points are sampled across the (user-specified) cutout along the x and y dimension?  Does this make any implict assumption about it being a square vs. an extreme rectangle?  E.g., does this work for a requested cutout that is 1 pixel in x-dimension and 50 pixels in the y-dimension?

scfleming: The fact that the while loops over both the x- and y-dimensions with a single index counter ("i") is why I wonder if it works as expected for rectangles.  If the dimension is not a square, does this technique require looping over two different index counters to account for the spacing being different between the two dimensions?  I'd have to try it out myself with a stand-alone script for sure, but wanted to document my thought on this here.

astrocut/cube_cut.py
```diff
@@ -272,17 +272,26 @@ def _fit_cutout_wcs(self, cutout_wcs, cutout_shape):
...
+        pix_inds = np.array(list(product(list(range(0,x,i)), list(range(0,y,i)))))
         world_pix = SkyCoord(cutout_wcs.all_pix2world(pix_inds, 0), unit='deg')
 
         # Getting the fit WCS
         linear_wcs = fit_wcs_from_points([pix_inds[:, 0], pix_inds[:, 1]], world_pix, proj_point=self.center_coord)
 
         self.cutout_wcs = linear_wcs
 
-        # Checking the fit
+        # Checking the fit (we want to use all of the pixels for this)
```
scfleming, line 290: What is being checked here?  Is "dists" used anywhere else?

ceb8: Nope, dists is the angular distance between the true skycoords as calculated with the original wcs and the approximate skycoords as calculated with the linear wcs.

ceb8: This is meant to be a measure of how good the approximate wcs is. It's very good for small cutouts and gets progressively worse for larger cutouts.

astrocut/cube_cut.py
ceb8: I believe this is working as intended and I will certainly add a comment. The goal is to create an evenly spaced mesh of no more than 100 pixel coordinates to use in the fit. In practice I do this by taking the ith value in each row/column.

This is what  happens for a 5x7 cutout with i = 2:
```
xoxoxox
ooooooo
xoxoxox
ooooooo
xoxoxox
```

Where x denotes the indexes that will be used in the fit.

Does that make sense?

scfleming: Yes it does, in fact this exact thing, including the tic-tac-toe board, would be a great thing to add as an in-line comment block.

ceb8: Will do!

ceb8:
Here are plots for four different parts of the sky: near the pole, near the equator, over the 0->360 crossover, and a random part of the sky I've been coding off of. I tested a variety of shapes/sizes from 5x5 up to 200x200 (the maximum cutout size the API allows). I also testes cutouts in the corner of the FFI, near the center, and halfway between.

At the 0->360 crossover:
354_66_100_plot

Near the equator:
125_7_100_plot

Near the pole:
104_-74_100_plot

Random sky position:
259_36_100_plot

Note how mostly the error tracks the size of the cutout, increasing slightly faster when using a maximum of 100 points, but there are exceptions, interestingly in both directions where either the new fit is better or the old one is. 

Also, while making these plots I noticed some issues with the crpix values sometimes being really off. Interestingly this did not correspond to higher error fits, and also was a problem for both the new and old fits. I switched to using "center" for the center coordinate rather than giving a specific SkyCoord and that seems to have solved the problem. I originally did not do that because it failed at the 0->360 degree crossover, a bug which appears to have been fixed.  The fact that giving a SkyCoord as the center value sometimes gives anomalous results is a bug in the fitting code that should be resolved in astropy, however as using "center" does work, that does not need to be fixed for this PR. (I plan to explore this further and put in an issue/PR with astropy).

ceb8:
OK I did more digging and made some more changes and improved the fit in a number of ways:
- Created the grid of pixels centered on the center pixel and made sure the the edge pixels are included
  This makes the output naxis values correct and, makes the input grid symmetrical and even which seems to make the fit go faster
- Added bounds in the wcs fitting code so that the CRPIX value must be on the cutout
  This takes care of the oddly bad fits I was sometimes getting
- Added 1 to the pixel shape in the wcs fitting code
  This code was 1 off because it was subracting pixel indices rather than lengths

The wcs fitting updates will go into a PR for astropy.

Here are the updated plots for the different parts of the sky, where we can see that the fits are actually improved through these changes.

0->360 crossover
354_66_100_plot

Near the equator:
125_7_100_plot

Near the pole:
104_-74_100_plot

Random sky position:
259_36_100_plot

PR #31, opened 2020-05-07
Replace drop_after argument

ceb8:
Older drizzle files (e.g. CANDELS) have SIP distortion keywords in their headers even though the drizzle process removes the need for them.  This causes problems when loading and using the associated WCS information. The current solution is to offer an option to drop all keywords after the first instance of a specific keyword.  This not optimal. 

This PR handles that issue by catching the warning that astropy issues when it finds inconstant WCS keywords, assumes that the CTYPE keywords are correct and drops all SIP coefficients in this case. Consequently the `drop_after` keyword is deprecated.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,7 @@
 ----------------
 - Update wcs fitting to match Astropy (and use Astropy when available) [#29]
 - Limit the number of pixels used for WCS fitting to 100 [#30]
+- Deprecate drop_after and handle inconsistant wcs keywords automatically [#31]
 
 
 0.5 (2020-01-13)
--- a/astrocut/cutouts.py
+++ b/astrocut/cutouts.py
@@ -15,7 +15,7 @@ from astropy import units as u
 from astropy.io import fits
 from astropy.coordinates import SkyCoord
 from astropy import wcs
-from astropy.utils.decorators import deprecated_renamed_argument
+from astropy.utils.exceptions import AstropyDeprecationWarning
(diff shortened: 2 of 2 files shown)


dr-rodriguez:
Overall, this looks very good. Here are a few suggestions:

You can avoid the nested loop in `remove_sip_coefficients` by doing something like
```
for temp in product(["A", "B"], ["", "P"], [0, 1, 2, 3], [0, 1, 2, 3]):
    lets = ''.join(temp[:2])
    print(lets)
    print('{}_{}_{}'.format(lets, temp[2], temp[3]))
```
That does mean that `lets` gets repeated values, though, so it's not a clear improvement.

Also, you could consider replacing the use of try..except with something like:
```
if lets in hdu_header.keys():
    del hdu_header["{}_ORDER".format(lets)]
```
I don't know if either of these changes would be better, though, so I'll leave that up to your discretion.

The logging handling is clever. Could it be possible that someone overwrites the logger and removes all handlers prior to calling astrocut? I don't think this would every happen in practice; and it's possible logger always has at least one handler.

Finally, for lines 269-272, why not avoid the if..else statement altogether with something like: 
```
hdu_header.update(cutout_wcs.to_header(relax=not no_sip))
```

ceb8:
@dr-rodriguez I made the log capture code robust against the issue you noted. Which was distressingly easy to trigger.

I did not completely remove the loops the way you suggested, but did at key checking in place of the try/catch syntax as I think it is neater.

PR #33, opened 2020-05-20
Changing memmap mode

ceb8:
This PR changes the memmap mode when reading a fits file from `readonly` to `denywrite,` the difference is that `readonly` map to memmap access mode "ACCESS_COPY" while `denywrite` maps to "ACCESS_READ." For our purposes the issue with "ACCESS_COPY" is that it requires more memory, which can cause problems for larger files.


ceb8:
@dr-rodriguez  Thanks! Yup, test suite is only working on very small files.

PR #40, opened 2021-03-24
Adding moving target cutouts

ceb8:
This PR supports moving target TESS cutouts.

A number of support functions have been added, the two most important of which are:

- **path_to_footprints** : This function takes in a path (list of skycoords) and a WCS object and returns a list of footprints of the form `{'center_coord': ~astropy.coordinate.SkyCoord, 'size': [ny,nx]}` suitable for using as arguments for a cutout function (most importantly tesscut).
- **center_on_path**: This function takes a path (list of skycoords and times) and a set of cutout files (as from tesscut) and returns a new target pixel file that is a cutouts centered on the moving target describes by the given path.

The basic work flow for moving target tesscut is:
1. Use JPHorizons to make the path and figure out what FFIs are in play
2. Use `path_to_footprints` to get arguments to give to cube_cut function
3. Perform one or more normal tess cutouts
4. Use `center_on_path` to create one target pixel file centered on the moving target from the tess cutout(s) made in the previous step.

\*Note: While this is complete code there are some improvements to still be made, noted as "to-dos" within the code. Additionally while all the new code has doc strings, it hasn't yet been added to the sphinx docs. Both of these things will be part of a future PR.


astrocut/cutout_processing.py
```diff
@@ -0,0 +1,429 @@
...
+    # Removing any coordinates outside of the img wcs
+    valid_locs = ((x >= 0) & (x < img_wcs.array_shape[0])) & ((y >= 0) & (y < img_wcs.array_shape[1]))
+    x = x[valid_locs]
+    y = y[valid_locs]
+    
+    bounds_list = _get_bounds(x, y, size)
+    
+    combined_bounds = []
+    cur_bounds = None
+    for bounds in bounds_list:
```
dr-rodriguez, line 159: This for loop took me quite a while to parse. Maybe some comments throughout it would help? 

I wonder: to avoid the if statement could you perhaps do something like:
```
cur_bounds = bounds_list[0]
for bounds in bounds_list[1:]:
    new_bounds = _combine_bounds(cur_bounds, bounds)
    etc etc
```

astrocut/cutout_processing.py
```diff
@@ -0,0 +1,429 @@
...
+    Parameters
+    ----------
+    path : `~astropy.table.Table`
+        Table (or similar) object with columns "time," containing `~astropy.time.Time`
+        objects and "position," containing `~astropy.coordinate.Skycoord` objects.
+    size : array
+        Size in pixels of the cutout rectangle centered on the path locations ther will be 
+        returned. Formatted as [ny,nx]
+    cutout_fles : list
+        List of strings, Target Pixel File paths that the path crosses.
```
dr-rodriguez, line 353: Isn't the input list of files is the Full Frame Image files?

dr-rodriguez: Ah, never mind. I see now that this is already working on the cutout files

astrocut/cutout_processing.py
```diff
@@ -0,0 +1,429 @@
...
+        
+        
+        cutout_table["TGT_X"] = cutout_table["x"] - cutout_table["bounds"][:, 0, 0]
+        cutout_table["TGT_Y"] = cutout_table["y"] - cutout_table["bounds"][:, 1, 0]
+        
+        positions = cutout_wcs.pixel_to_world(cutout_table["x"], cutout_table["y"])
+        cutout_table["TGT_RA"] = positions.ra.value
+        cutout_table["TGT_DEC"] = positions.dec.value
+        
+        cutout_table["bounds"] = [(slice(*y), slice(*x)) for x, y in cutout_table["bounds"]]
```
dr-rodriguez, line 246: In most cases you have bounds being [[x],[y]]. I want to make sure here you do mean to have `slice(*y)` first. 

Some extra comments around this for loop could also help.

astrocut/cutout_processing.py
```diff
@@ -0,0 +1,429 @@
...
+    nx = bounds[0, 1]-bounds[0, 0]
+    ny = bounds[1, 1]-bounds[1, 0]
+    x = nx/2 + bounds[0, 0]
+    y = ny/2 + bounds[1, 0]
+    return {"coordinates": img_wcs.pixel_to_world(x, y),
+            "size": (ny, nx)}
+
+
+# TODO: Put this in utils 
+def path_to_footprints(path, size, img_wcs, max_pixels=10000):
```
dr-rodriguez, line 121: I did not see this function used anywhere except in the tests. I guess that is used elsewhere and that's why you have the todo to move it to utils?

astrocut/cutout_processing.py
ceb8: Nope, this expects normal cutout target pixel files.

astrocut/cutout_processing.py
ceb8: Yeah, this is correct, I think it's because of the difference in x,y vs y,x in fits vs python. 😛 I'll add some more comments.

astrocut/cutout_processing.py
ceb8: Yes, sorry, I should have been more clean in the PR comment. See my edited comment about the moving target tesscut workflow.

ceb8: TLDR: Moving target tesscut uses this function and `center_on_path` for the full workflow, but they are not dependent on each other individually.

dr-rodriguez (approved):
At a glance, this looks OK. 
A few parts of the code would benefit for a little bit more commenting as they were hard to follow.

PR #44, opened 2021-06-02
[WIP] Enable `CutoutFactory.cube_cut()` to use S3-hosted files

barentsen:
## Summary

This PR documents a working prototype which enables the `CutoutFactory.cube_cut()` method to read TESS cube files directly from AWS S3 storage, by-passing the need for users to download the large cubes locally. (TESS cubes amount to ~110 GB per CCD in Sector 35!)

Example use:

Screen Shot 2021-06-01 at 10 41 46 PM

The motivation for this change is to enable fast cloud-based analyses of TESS FFI data, i.e., this feature would enable users to obtain thousands of TESS cut-outs concurrently, limited only by the throughput of S3.

This PR is rough and not ready for review -- it lacks tests, comments, etc.  I also haven't investigated whether this change would be compatible with @ceb8's awesome recent work on moving cut-outs.  However I am happy to prepare a proper PR if we think this is a good idea!

## Technical details

This prototype defines two compatible classes, `LocalCubeFile` and `S3CubeFile`, which encapsulate two different data access patterns:
* `LocalCubeFile` uses `astropy.io.fits.open()` to access a cube as normal.
* `S3CubeFile` uses the `boto3` and `asyncio` libraries to asynchronously extract slices from the cube using `Byte-range` GET requests.  The number of requests scales with the vertical size of the cutout (`requests = n_rows + 3`), leveraging the smart columnar way in which the cubes are organized for this purpose.

The `S3CubeFile` will only work with astrocut-produced cubes at this time. In the future we can hopefully integrate cloud streaming features in a more generic way into `astropy.io.fits`, but that would be a much more expensive project.  See existing discussions at https://github.com/astropy/astropy/issues/3446 and https://github.com/jbcurtin/astro-cloud/issues/1.

## Inspiration

The idea to stream partial FITS files from the cloud is not mine and has been explored by many others before me, including:
* @astrofrog, cf. https://gist.github.com/astrofrog/404a5682311bae221fd5308c1498da94
* @pllim & @mustaric, cf. https://github.com/mustaric/lambda-tess-search/blob/master/sandbox/aws_remote_fits.py
* @jbcurtin, cf. https://github.com/jbcurtin/astro-cloud/
* and others, no doubt!

## Verification

<details>
  <summary>Click here to see the code snippet I used to verify that S3-based cutouts match local cutouts:</summary>

```python
from astrocut import CutoutFactory
path = "/Users/gb/tmp/tess-s0035-1-1-cube.fits"
CutoutFactory().cube_cut(path, "147 -7.5", 10, "local-cutout.fits")
uri = "s3://stpubdata/tess/public/mast/tess-s0035-1-1-cube.fits"
CutoutFactory().cube_cut(uri, "147 -7.5", 10, "s3-cutout.fits")

import lightkurve as lk
lk.read("local-cutout.fits").plot(title="Local Cube");
lk.read("s3-cutout.fits").plot(title="S3 Cube");
```
Screen Shot 2021-06-01 at 11 02 33 PM

</details>


astrocut/cube_cut.py
```diff
@@ -800,3 +810,183 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
...
+
+    @property
+    def table(self):
+        return self._read_table()
+
+    @lru_cache()
+    def _read_table(self):
+        """Returns the BinTableHDU for HDU #2."""
+        hdu1_data_size = np.product(self.shape) * self.itemsize
+        hdu2_offset = self.HDU1_DATA_OFFSET + 2880 * (1 + hdu1_data_size // 2880)
```
mfox22, line 918: Magic number 2880, should this be HDU1_HEADER_OFFSET?

mfox22 (commented):
Excited to see someone picking up this work!
@barentsen have you done any validation testing against the current TESScut web service/api or against the local package option?

Whoops I see where you did mention testing against local. Very nice.

jbcurtin:
The algorithm looks pretty complete. I actually tried to do this a second time, but with less group direction. Here is some more source, in-case you get lost like I did: https://github.com/jbcurtin/tess-enumerator

I concluded that some of the fits files were inconsistent relative to the header information. Might be wise to make sure the data-cubes are all properly formatted. Or they might be, and its my algorithm that is wrong 🤷

astrocut/cube_cut.py
barentsen: 2880 here refers to the fact that the FITS standard prescribed that *"each FITS structure shall consist of an integral number of FITS blocks, which are each 2880 bytes"*.  This deserves a comment in the code at minimum, and of course really ought to be defined as a `FITS_BLOCK_SIZE` constant instead of being hard-coded.

barentsen:
@mfox22 @jbcurtin Thank you for the early feedback!  I'll clean this up and turn this into a proper PR with tests/docstrings/validation/etc.

@jbcurtin I'll go ahead and verify that the TESSCut cubes all have a consistent structure.  Thanks for the word of caution!  In the case of the original TESS FFI files, it is definitely true that not all files have the same structure. Specifically, some "bad quality"-flagged FFI's do not include WCS header keywords, which alters the length of the header and affects the starting byte position of the image data. I hope that the TessCut cubes are more consistent.  (PS: It looks like the `stpubdata` bucket on S3 is free to access now, so no need to pay $1000 any longer!)

mustaric:
I'm so pleased to see this being done, it looks like a great addition to astrocut.  
I assume you have a commitment from MAST (and AWS) to continue hosting the TESScut cubes in the cloud. I remember this being the issue from implementing S3 cutouts in this way because of the extra data volume required.

falkben:
> @falkben Thank you for raising the valid questions about `nest_asyncio` above. I'm writing this separate comment to document my motivation for using `asyncio` and `nest_asyncio` for posterity. (Caveat: I'm not an expert on this topic and happy to be proven wrong.)
> 
>     * **Why use `asyncio`?** Cutting a square of (n_rows, n_columns) pixels from the TESS cubes hosted in AWS S3 requires n_rows byte-range requests[1](#user-content-fn-1-4d2c8666d6e70ce65e811367c36f60b1), because data for different rows are stored at very different byte locations.  Making multiple slow i/o requests naturally benefits from some form of concurrency.  I found that using coroutines (i.e. asyncio) for this problem is much faster than using threads or multiprocessing.  This makes sense because threads create more overhead, while multi-processing does not make sense because we are i/o-limited rather than cpu-limited.
> 
>     * **Why use `nest_asyncio`?** A feature of Python's built-in coroutine module (asyncio) is that it does not allow event loops to be started in an environment that is already running an event loop.  This makes sense at face value because nested event loops may lead to confusing bugs and wasted resources.  Unfortunately, this means that different code would be required to run the cube cutouts in an environment that is already running an event loop (e.g. Jupyter Notebook) versus an environment without event loop (e.g. a script running in a standard interpreter).  `nest_asyncio.apply()` removes the need for different code by patching `asyncio` to allow for nested event loops to exist.
> 
>     * **Does `nest_asyncio` have side-effects?**  Yes, it changes the behavior of asyncio for all code that is running in the same kernel.  I haven't encountered a situation where this is an issue, and the internet suggests that `nest_asyncio` is being used at scale, however it would be much nicer if this were a standard feature of asyncio.  I have not tried it in conjunction with fastapi.
> 
>     * **Can we get rid of `nest_asyncio`?**  Yes.  In this case, we would have to instruct users to call `nest_asyncio.apply()` when attempting to create S3-based cutouts in a Notebook environment.   Fortunately, we could include this instruction in the error message that is raised when it is omitted, so it's a reasonable solution.
> 
> 
> ## Footnotes
> 
>     1. Note: the HTTP protocol allows retrieving multiple byte ranges with a single GET request, however AWS S3 does not support this and requires a separate GET request for each byte-range. [↩](#user-content-fnref-1-4d2c8666d6e70ce65e811367c36f60b1)

Thanks for the very thorough explanation.

My first thought is to try to avoid nested event loops entirely. In order to grab/create the event loop and just run things in the main event loop, can we do something like:

```python
try:
    # get the event loop if one is already running
    loop = asyncio.get_running_loop()
except RuntimeError:
    # if one is not running, create it
    loop = asyncio.get_event_loop()
```

The other thing I'm noticing is that we're doing multiple calls to `asyncio.run()`

My understanding is that `asyncio.run()` should be used as an entry point. In fact, in my above code snippet, the except block might be a good place for an `asyncio.run()` call. 

From the [docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.run):

> This function cannot be called when another asyncio event loop is running in the same thread.

and

> This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.

which makes me think this might be the source of the current need for `nest_asyncio`, as each time we call `asyncio.run()` we're always creating a new event loop and closing it at the end.

barentsen:
> The other thing I'm noticing is that we're doing multiple calls to asyncio.run()
> My understanding is that asyncio.run() should be used as an entry point. 

Interesting!  Thank you for pointing this out.  I am going to read the docs more carefully and attempt a refactor with a single call to `run()`.

scfleming:
> I get the same warning (`VerifyWarning: Card is too long`) when obtaining local cutouts using the `main` branch of this repository, so I'm fairly confident the warning is not triggered by any of the changes in this PR. My favorite way to debug a warning like this is to run a script or test using `python script.py -W error --pdb`, which turns warnings into exceptions and automatically opens the `pdb` debugger at the point in the code where the warning was triggered. I will investigate and report back!

RE: this error, this is probably astropy letting you know one or more of the FITS header Cards has a comment that pushes it above 80 characters, and thus will be truncated. This is run as part of its verification of FITS files that it creates, one way I like to check that (if primary header) is to "more" the FITS file in a terminal that is exactly 80 or 160 characters long (so it displays nicely) and then see if any of the comments are cut off. This doesn't work well of course if it's an extension, but you can also then just read in the header keyword comments and print them out in a Python script to check if any are obviously truncated...

PR #45, opened 2021-06-10
Hap cutout work

ceb8:
PR for the HAP project. WIP because it needs the full combine functionality in place.


astrocut/tests/test_cutout_processing.py
```diff
@@ -254,11 +256,79 @@ def test_center_on_path(tmpdir):
     assert hdu[0].header["DATE"] == Time.now().to_value('iso', subfmt='date')
     assert hdu[0].header["OBJECT"] == ""
     hdu.close()
-    
 
-    
 
-    
+def test_default_combine():
```
dr-rodriguez, line 261: Some more description for this test (eg, that build_default_combine_function is creating masks of the input array) can help others understand why some of the assertions work the way they do.

dr-rodriguez (approved):
Skimmed through the code and it seems OK to me. Looks like a lot of functions were moved to utils and that's a lot of the changes. I spent most of the time looking at cutout_processing and its tests to try to understand the logic behind combining cutouts.

ceb8:
@dr-rodriguez  Yes, as part of this PR I moved a lot of functions around. I've added a docstring to explain the combine function tests.

PR #52, opened 2021-09-29
Addition of check for cutout existing before continuing

phicc:
With the previous updates, the check confirming that the file exists was removed. Unfortunately, there appears to be edge cases where the file may be empty and this causes an index error. This work adds a check prior to looping through the cutout file dictionary. Also correcting a minor spelling mistake.

FYI: Due to mission freezes, we will have to coordinate the timing of the merging and promotion to PyPi.

### DIFF
--- a/astrocut/cutouts.py
+++ b/astrocut/cutouts.py
@@ -559,7 +559,7 @@ def img_cut(input_files, coordinates, cutout_size, stretch='asinh', minmax_perce
 
     # If no cutouts contain data, raise exception
     if not cutout_hdu_dict:
-        raise InvalidQueryError("Cutout contains to data! (Check image footprint.)")
+        raise InvalidQueryError("Cutout contains no data! (Check image footprint.)")
 
     # Make sure the output directory exists
     if not os.path.exists(output_dir):
@@ -593,6 +593,11 @@ def img_cut(input_files, coordinates, cutout_size, stretch='asinh', minmax_perce
  
         cutout_path = []
         for fle in input_files:
+
+            if cutout_hdu_dict.get(fle) is None:
+                warnings.warn("Cutout of {} contains no data and will not be written.".format(fle),
+                              DataWarning)
+                continue
 
             for i, cutout in enumerate(cutout_hdu_dict.get(fle)):
 


dr-rodriguez (approved):
Looks reasonable to me

dr-rodriguez:
> FYI: Due to mission freezes, we will have to coordinate the timing of the merging and promotion to PyPi.

I take it our applications using astrocut don't have a pinned version then?

falkben:
> > FYI: Due to mission freezes, we will have to coordinate the timing of the merging and promotion to PyPi.
> 
> I take it our applications using astrocut don't have a pinned version then?

They do. It's not clear to me why astrocut is subject to the freeze, but that's the info that we currently have.

astrocut/cutouts.py
```diff
@@ -594,6 +594,11 @@ def img_cut(input_files, coordinates, cutout_size, stretch='asinh', minmax_perce
         cutout_path = []
         for fle in input_files:
 
+            if cutout_hdu_dict.get(fle) is None:
```
falkben, line 597: i think this check is fine, but seems like there might be something going wrong here: https://github.com/spacetelescope/astrocut/blob/56168752cfcd49ff948ab3443f11dbf644246a06/astrocut/cutouts.py#L554

it's just not clear to me how this could be `None`

is there a check that we should but in there as well?

is it possible to add a test that failed before and will now pass with this additional check?

scfleming: Ah, that does change the game if so.  I'm meeting with @phicc tomorrow to talk through further, stay tuned!  Just so I know, this would be z.MAST only that would be impacted by Astrocut changes that relates to JWST?  No Portal, etc. reliance?

scfleming:
> > > FYI: Due to mission freezes, we will have to coordinate the timing of the merging and promotion to PyPi.
> > 
> > 
> > I take it our applications using astrocut don't have a pinned version then?
> 
> They do. It's not clear to me why astrocut is subject to the freeze, but that's the info that we currently have.

Ah, that does change the game if so. I'm meeting with @phicc tomorrow to talk through further, stay tuned! Just so I know, this would be z.MAST only that would be impacted by Astrocut changes that relates to JWST? No Portal, etc. reliance?

astrocut/cutouts.py
seweissman: This is wrapped in an if, so it seems like this could be None if `cutout` is all zeros. It seems like setting `cutout_hdu_dict` to `default_dict(list)` would maybe fix this issue and would be consistent with `cutout_hdu_dict.get(in_fle, [])`.

https://github.com/spacetelescope/astrocut/blob/56168752cfcd49ff948ab3443f11dbf644246a06/astrocut/cutouts.py#L553-L554

phicc:
> > > > FYI: Due to mission freezes, we will have to coordinate the timing of the merging and promotion to PyPi.
> > > 
> > > 
> > > I take it our applications using astrocut don't have a pinned version then?
> > 
> > 
> > They do. It's not clear to me why astrocut is subject to the freeze, but that's the info that we currently have.
> 
> Ah, that does change the game if so. I'm meeting with @phicc tomorrow to talk through further, stay tuned! Just so I know, this would be z.MAST only that would be impacted by Astrocut changes that relates to JWST? No Portal, etc. reliance?

The only applications that would be impacted are ZCut and ZMAST (which makes API calls to ZCut). We do pin our versions in ZCut.

falkben (approved):

PR #56, opened 2022-03-15
fixing broken tests from GitHub Actions CI run

jaymedina:
Fixing deprecated scripts that are causing CI tests to break. Changes include:

- [x] run `apt-get update` before `apt-get install` step in CI workflow
- [x] fix `ERROR: the documentation requires the sphinx-astropy package to be installed`

Tests are passing except for the allowed failure for python 3.7, so I will mark this PR as ready for review. Thanks in advance!

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -5,6 +5,7 @@ on:
     branches:
     - main
     tags:
+    - '*'
   pull_request:
 
 jobs:
@@ -56,9 +57,12 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
+    - name: Updating apt-get before install
+      if: startsWith(matrix.os, 'ubuntu')
+      run: sudo apt-get update 
     - name: Install language-pack-de and tzdata
       if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get install language-pack-de tzdata
+      run: sudo apt-get --fix-missing install language-pack-de tzdata
     - name: Install Python dependencies
       run: python -m pip install --upgrade tox codecov
     - name: Run tests
@@ -96,9 +100,12 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
+    - name: Updating apt-get before install
+      if: startsWith(matrix.os, 'ubuntu')
+      run: sudo apt-get update 
     - name: Install language-pack-de and tzdata
       if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get install language-pack-de tzdata
+      run: sudo apt-get --fix-missing install language-pack-de tzdata
     - name: Install Python dependencies
       run: python -m pip install --upgrade tox codecov
     - name: Run tests
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -57,12 +57,11 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python }}
-    - name: Updating apt-get before install
-      if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get update 
     - name: Install language-pack-de and tzdata
       if: startsWith(matrix.os, 'ubuntu')
-      run: sudo apt-get --fix-missing install language-pack-de tzdata
+      run: | 
+        sudo apt-get update
+        sudo apt-get install language-pack-de tzdata
     - name: Install Python dependencies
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_workflows.yml
```diff
@@ -5,11 +5,12 @@ on:
     branches:
     - main
     tags:
+    - '*'
   pull_request:
 
 jobs:
   tests:
-    name: ${{ matrix.name }}
+    name: Name ${{ matrix.name }}
```
pllim, line 13:
```suggestion
    name: ${{ matrix.name }}
```

setup.cfg
```diff
@@ -41,7 +41,7 @@ testpaths = "astrocut" "docs"
 astropy_header = true
 doctest_plus = enabled
 text_file_format = rst
-addopts = --doctest-rst
+addopts = --doctest-modules
```
falkben, line 44: This seems like it might have different behavior. `--doctest-rst` uses pytest-doctestplus to test the code in the rst doc files I believe. Whereas `doctest-modules` tests the package's python modules and their docstrings

setup.cfg
```diff
@@ -19,6 +19,7 @@ install_requires =
     astropy >=3.0
     scipy
     Pillow
+    sphinx-astropy
```
falkben, line 22: why was this moved from the docs dependencies into the main dependencies? seems like it's not a dependency that should be need to run astrocut

falkben (changes requested):
We remove python 3.6 and add 3.9 and 3.10. could be done in a separate PR if that's easier. 3.6 is EOL.

jaymedina:
Attempting to run `apt-get update` and `install` under one step results in the following error, so I'm going to revert: 

`[Invalid workflow file: .github/workflows/ci_workflows.yml#L63](https://github.com/spacetelescope/astrocut/actions/runs/1999272225/workflow)
The workflow is not valid. .github/workflows/ci_workflows.yml (Line: 63, Col: 7): 'run' is already defined .github/workflows/ci_workflows.yml (Line: 104, Col: 7): 'run' is already defined`

See log: https://github.com/spacetelescope/astrocut/actions/runs/1999272225/workflow

setup.cfg
falkben: `sphinx-astropy` was previously defined by the `docs` dependencies here: https://github.com/spacetelescope/astrocut/blob/main/setup.cfg#L33 

I think the issue might be that `alldeps` in tox is not pulling in the docs dependencies.

these _should_ be installed by tox under specific test configurations when testing the docs. It seems like they are not. With my limited experience with `tox` i'm not sure what the easiest way is to fix it.

We also don't even have an `all` configuration in `setup.cfg` under `extras_require`  See that we refer to `all` here in tox.ini: https://github.com/spacetelescope/astrocut/blob/main/tox.ini#L61 but there's no all config in extras_require

One possible fix would be to include `docs` for `alldeps`. I've not tested this but it seems like it might work. Change:

`alldeps: all`

to 

`alldeps: docs`

here: https://github.com/spacetelescope/astrocut/blob/main/tox.ini#L61

tox.ini
```diff
@@ -9,6 +9,7 @@ envlist =
 requires =
     setuptools >= 30.3.0
     pip >= 19.3.1
+    sphinx-astropy >= 1.7.0
```
falkben, line 12: this should not be a requirement of tox

falkben (changes requested):

setup.cfg
jaymedina: OK will do that now. Quick q: Originally the `alldeps` was set to `all` - if the `all` doesn't exist, what was it doing? I imagine it would break with some sort of message saying `all` is not a thing

tox.ini
jaymedina: Removed

setup.cfg
falkben: i have no idea.

i may also have this backwards and it's "setting" to `all` was actually only a description of "alldeps". which would still not make sense, because i don't know how it was pulling in the docs dependencies there, either.

I think this will require some investigation into tox configuration and `extras_require`

maybe @pllim can give us a few ideas?

falkben: Seems like this suggestion got tox to install `sphinx-astropy`.

still a mystery to me how this was ever working before, when we had no `all` config.

pllim: You only use alldeps if you have `all` defined. Otherwise, it means nothing to your CI.

Without knowing the history of your code, I cannot tell you why it was working before. "alldeps" does not automatically grab "all". "alldeps" is just a name, and you can make it install whatever you want, really.

falkben: Thanks! This was my intuition, as well.

PR #59, opened 2022-04-12
Make cubes out of TICA FFIs

jaymedina:
Implements a class based off the existing `CubeFactory` class that can generate cubes out of TICA FFIs. 

Note: This is one of the core four projects for MAST Q3 and must be implemented before June 30. 

- [x] Implement TICACubeFactory
- [x] Verify file structural similarity between TICA and TESS Cubes
- [x] Verify image similarity between TICA and TESS Cubes
- [x] Add functionality to append incoming FFIs to existing cubes (starting in Sept we get new ones every week, EM1 we get them ever 2 weeks...?)
- [x] Fix truncated string issue: `StringTruncateWarning: truncated right side string(s) longer than 9 character(s) during assignment
  self.info_table[kwd][i] = ffi_data[0].header.get(kwd, nulval)`
- [x] First test that the duplicate columns issue is still an issue in delivery 1 of future sectors - so far I think I've only seen it in sector 27. ~Fix duplicate columns issue in build_info_table(). For some reason there are duplicate columns in the order 1 deliveries for keywords: TESS_X, TESS_Y, TESS_VX, TESS_VY etc~
- [x] Code cleanup: Make sure class variables are under _init_ so they get reset for every run. There was an issue with cube_shape being too small, that would get fixed when I reinitialized the class, investigate that. 
- [x] Unit test(s)

UPDATE: After investigating TruncatedString warning, I found the source of the issue being under `_build_info_table` where the default length of a string keyword value is set based on the length of the string keyword value in the designated `template_file` . In the event that the string kwd value in one FFI is longer than the string kwd value in the template file, a stringTruncated Warning will pop up. I fixed the default value and compared the generated `info_table` with the `info_table` before fixing, and it doesn't seem to clip off the kwd value the way I expected it would. So I'm going to leave this error alone and avoid fiddling with the string column length in case that creates a memory issue. The Warning will be noted here for future reference in case I need to revisit. 

Screenshot below shows the `DATASUM` values for 2 FFIs in an info table where one value is 9 characters and the other is 10. This is the string column causing the Warnings. 
image

----
----
# TICA vs SPOC Cube Analysis 

### HDU Lists for SPOC and TICA cubes (respectively)

image

----

#### Sector 27, Camera 1, CCD 2

image

#### Normalized (`Array/MEDIAN(Array)`)

image

#### Using this star as a reference point to verify alignment between two FFIs within the respective cubes

image

#### Ratio image (TESS/TICA) between exact cutouts of that star for both SPOC and TICA

image

#### Verifying time-cadence between SPOC and TESS. Start time between consecutive FFIs lapsed for 10 minutes. Consistent with the EM1 protocol. 

image

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -85,11 +85,6 @@ jobs:
             python: 3.7
             toxenv: py37-test-devdeps
 
-          - name: (Allowed Failure) Python 3.8 with all optional dependencies (Windows)
-             os: windows-latest
-             python: 3.8
-             toxenv: py38-test-alldeps
-
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -85,6 +85,11 @@ jobs:
             python: 3.7
             toxenv: py37-test-devdeps
 
+          - name: Python 3.8 with all optional dependencies (Windows)
+             os: windows-latest
+             python: 3.8
+             toxenv: py38-test-alldeps
+
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -28,6 +28,11 @@ jobs:
             toxenv: py38-test-alldeps
             toxargs: -v --develop
 
+          # Commenting out the test below due to outdated ubuntu OS 
+          # (16.04) which is no longer supported, 
+          # and Python 3.6 is no longer being supported. 
+          # TODO: Might be worth keeping this test for users who 
+          # are still on 3.6 and just update the os to ubuntu-18.04
           #- name: Python 3.6 with oldest supported version of all dependencies
           #  os: ubuntu-16.04
           #  python: 3.6
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -85,7 +85,7 @@ jobs:
             python: 3.7
             toxenv: py37-test-devdeps
 
-          - name: Python 3.8 with all optional dependencies (Windows)
+          - name: (Allowed Failure) Python 3.8 with all optional dependencies (Windows)
              os: windows-latest
              python: 3.8
              toxenv: py38-test-alldeps
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
     - main
-    - tica-make-cubes
     tags:
     - '*'
   pull_request:
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
     - main
+    - tica-make-cubes
     tags:
     - '*'
   pull_request:
@@ -37,11 +38,6 @@ jobs:
             os: ubuntu-latest
             python: 3.7
             toxenv: py37-test-alldeps-numpy121-cov
-
-          - name: Python 3.8 with all optional dependencies (Windows)
-            os: windows-latest
-            python: 3.8
-            toxenv: py38-test-alldeps
             
           - name: Python 3.7 with all optional dependencies (MacOS X)
             os: macos-latest
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -28,10 +28,10 @@ jobs:
             toxenv: py38-test-alldeps
             toxargs: -v --develop
 
-          - name: Python 3.6 with oldest supported version of all dependencies
-            os: ubuntu-16.04
-            python: 3.6
-            toxenv: py36-test-oldestdeps   
+          #- name: Python 3.6 with oldest supported version of all dependencies
+          #  os: ubuntu-16.04
+          #  python: 3.6
+          #  toxenv: py36-test-oldestdeps   
             
           - name: Python 3.7 with numpy 1.21 and full coverage
             os: ubuntu-latest
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -635,7 +635,7 @@ class TicaCubeFactory():
         self.cube_file = cube_file
 
         if verbose:
-            print(f'Working on cube file: {cube_file}')
+            print(f'Updating cube file: {cube_file}')
 
         # Ensure that none of the files in file_list are in the cube already, to avoid duplicates
         in_cube = list(fits.getdata(self.cube_file, 2)['FFI_FILE'])
@@ -647,7 +647,7 @@ class TicaCubeFactory():
         self._configure_cube(file_list)
 
         if verbose:
-            print(f"Cube will be appended in {self.num_blocks} blocks of {self.block_size} rows each.")
+            print(f"FFIs will be appended in {self.num_blocks} blocks of {self.block_size} rows each.")
         
         # Preparing to update the info table to hold image header keywords of the new FFIs
         self.info_table = fits.getdata(self.cube_file, 2)
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -555,16 +555,24 @@ class TicaCubeFactory():
             if verbose:
                 st = time()
 
+            # In this section we will take the SCI data from self.file_list above
+            # and "paste" a cutout of the full SCI array into a 4d array called 
+            # sub_cube. We iterate this process until the full SCI array for each 
+            # FFI is copied onto the cube. Usually the sub_cube is ~1/3 the size of 
+            # the full SCI array, so there are 3 iterations. 
+            # For TICA, the SCI data exists in the 0th extension. 
             with fits.open(fle, mode='denywrite', memmap=True) as ffi_data:
 
-                # add the image and info to the arrays
+                # Add the image and info to the arrays
                 sub_cube[:, :, i, 0] = ffi_data[0].data[start_row:end_row, :]
 
                 del ffi_data[0].data
                
-                if fill_info_table:  # Also save the header info in the info table
+                # Also save the header info in the info table
+                if fill_info_table:
 
-                    for kwd in self.info_table.columns:  # Iterate over every keyword in the TICA FFI primary header
+                    # Iterate over every keyword in the TICA FFI primary header
+                    for kwd in self.info_table.columns:
                         if kwd == "FFI_FILE":
                             self.info_table[kwd][i] = os.path.basename(fle)
                             
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -454,8 +454,8 @@ class TicaCubeFactory():
         else:
             with fits.open(self.cube_file, mode='update', memmap=True) as cube_hdu:
                 header = cube_hdu[0].header 
-                header['HISTORY'] = f'Updated on {str(date.today())} with new FFI delivery.' 
-                header['HISTORY'] = f'First FFI is {str(os.path.basename(self.file_list[0]))}'
+                #header['HISTORY'] = f'Updated on {str(date.today())} with new FFI delivery.' 
+                #header['HISTORY'] = f'First FFI is {str(os.path.basename(self.file_list[0]))}'
 
         # Adding the keywords from the last file
         with fits.open(self.file_list[-1], mode='denywrite', memmap=True) as last_file:
@@ -634,8 +634,6 @@ class TicaCubeFactory():
         if not self.update:
             cube_hdu[1].data[start_row:end_row, :, :, :] = sub_cube
         else:
-            print(self.cube_append[start_row:end_row, :, :, :].shape)
-            print(sub_cube.shape)
             self.cube_append[start_row:end_row, :, :, :] = sub_cube
 
         if (version_info <= (3, 8)) or (platform == "win32"):
@@ -770,9 +768,9 @@ class TicaCubeFactory():
             hdul[1].data = new_cube
 
         # Add the info table to the cube file
-        self._write_info_table()
-        if verbose:
-            print(f"Total time elapsed: {(time() - startTime)/60:.2f} min")
+        #self._write_info_table()
+        #if verbose:
+        #    print(f"Total time elapsed: {(time() - startTime)/60:.2f} min")
 
         return self.cube_file
 
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -730,7 +730,7 @@ class TicaCubeFactory():
             cube_hdus.append(table_hdu)
 
 
-    def update_cube(self, file_list, cube_file, verbose=True):
+    def update_cube(self, file_list, cube_file, sector=None, verbose=True):
         """ Updates an existing cube file if one has already been made and a new delivery is being appended to it. 
         Same functionality as make_cube(...), but working on an already existing file rather than building a new one. 
         This function will: 
@@ -770,7 +770,9 @@ class TicaCubeFactory():
                     print(f'FFI {file} is already in the cube. Removing it from ``file_list``.')
                 file_list.remove(file)
         
-        self._configure_cube(file_list)
+        # Set up the basic cube parameters
+        sector = (sector, "Observing sector")
+        self._configure_cube(file_list, sector=sector)
 
         if verbose:
             print(f"FFIs will be appended in {self.num_blocks} blocks of {self.block_size} rows each.")
@@ -839,7 +841,7 @@ class TicaCubeFactory():
         return self.cube_file
 
 
-    def make_cube(self, file_list, cube_file='img-cube.fits', verbose=True, max_memory=50):
+    def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memory=50, verbose=True):
 
         if verbose:
             startTime = time()
@@ -847,7 +849,8 @@ class TicaCubeFactory():
         self.max_memory = max_memory
 
         # Set up the basic cube parameters
-        self._configure_cube(file_list)
+        sector = (sector, "Observing sector")
+        self._configure_cube(file_list, sector=sector)
 
         if verbose:
             print("Using {} to initialize the image header table.".format(os.path.basename(self.template_file)))
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -427,6 +427,7 @@ class TicaCubeFactory():
         # If it's a new cube, the shape is (nRows, nCols, nImages, 2)
         if not self.update:
             self.cube_shape = (image_shape[0], image_shape[1], len(self.file_list), 2)
+            
         # Else, if it's an update to an existing cube, the shape is (nRows, nCols, nImages + nNewImages, 2)
         else: 
             self.cube_shape = self.cube_append.shape
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -415,7 +415,16 @@ class TicaCubeFactory():
         slice_size = image_shape[1] * len(self.file_list) * 2 * 4  # in bytes (float32)
         max_block_size = int((self.max_memory * 1e9)//slice_size)
         
+        print('SLICE SIZE, MAX BLOCK SIZE, IMAGE SHAPE')
+        print(slice_size)
+        print(max_block_size)
+        print(image_shape[0])
+    
         self.num_blocks = int(image_shape[0]/max_block_size + 1)
+
+        print('NUM BLOCKS')
+        print(self.num_blocks)
+
         self.block_size = int(image_shape[0]/self.num_blocks + 1)
         self.cube_shape = (image_shape[0], image_shape[1], len(self.file_list), 2)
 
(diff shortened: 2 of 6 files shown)


jaymedina:
Comparison of first 5 images from TICA and TESS cubes, sector 27, camera 1, ccd 2: 
image

jaymedina:
Note to self: Make sure that TICA cube BinTable (ext2) has keywords from the FFI ext1 as well.

astrocut/make_cube.py
```diff
@@ -346,3 +348,545 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        # fits requires all blocks to be a multiple of 2880
+        cubesize_in_bytes = ((np.prod(self.cube_shape) * 4 + 2880 - 1) // 2880) * 2880
+        filelen = os.path.getsize(cube_file)
+        
+        with open(cube_file, 'r+b') as CUBE:
+            # What is this .seek?
+            # Looks like it's expanding the buffer size of the file?
+            CUBE.seek(filelen + cubesize_in_bytes - 1)
+            # b for buffer?
+            CUBE.write(b'\0')
```
jwfraustro, line 550: From what I understand of this, it's just seeking to the end of the `cube_file` and writing one additional null byte, `b'\0'` just being the the null character as a byte. I don't see precisely why it's being added yet, except maybe as a terminator, though from what I recall of the FITS standard only text column tables are null-terminated.

jaymedina: Interesting, yeah I'm not sure why that final byte is in there, it's something I'll investigate later and add to the documentation. Thanks for clarifying!

falkben (changes requested):
Can we use class inheritance here? It looks like most of the changes would just be in the class attributes. If we make those changes, I think we can make these changes much simpler and should be easier to maintain.

astrocut/make_cube.py
```diff
@@ -346,3 +349,496 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        Set up the cube primary header.
+        """
+
+        file_list = np.array(file_list)
+        image_shape = None
+        start_times = np.zeros(len(file_list))
+        for i, ffi in enumerate(file_list):
+            ffi_data = fits.open(ffi, mode='denywrite', memmap=True)
+            
+            start_times[i] = ffi_data[0].header.get('STARTTJD')
```
falkben, line 400:
```suggestion
            start_times[i] = ffi_data[0].header.get(self.time_keyword)
```

astrocut/make_cube.py
```diff
@@ -346,3 +349,496 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+
+            if image_shape is None:  # Only need to fill this once
+                image_shape = ffi_data[0].data.shape
```
falkben, line 403: Since the ffi_data order is diff. between TICA and normal, we might want to set these indices as class attributes.

(image data is 0 for TICA but 1 for the original ones)

astrocut/make_cube.py
```diff
@@ -346,3 +349,496 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        self.block_size = int(image_shape[0]/self.num_blocks + 1)
+        
+        # Determining cube shape: 
+        # If it's a new cube, the shape is (nRows, nCols, nImages, 2)
+        if not self.update:
+            self.cube_shape = (image_shape[0], image_shape[1], len(self.file_list), 2)
+            
+        # Else, if it's an update to an existing cube, the shape is (nRows, nCols, nImages + nNewImages, 2)
+        else: 
+            self.cube_shape = self.cube_append.shape
```
falkben, line 434: Just wondering why this is different compared to the default `CubeFactory`

In the original factory, it seems like cube_shape is always `(image_shape[0], image_shape[1], len(self.file_list), 2)`

do those no receive updates?

astrocut/make_cube.py
```diff
@@ -346,3 +349,496 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+                # Adding the extra keywords passed in
+                for kwd, value in extra_keywords.items():
+                    header[kwd] = (value[0], value[1])
+
+        # Otherwise we're updating an existing cube file
+        else:
+            with fits.open(self.cube_file, mode='update', memmap=True) as cube_hdu:
+                header = cube_hdu[0].header 
+                # header['HISTORY'] = f'Updated on {str(date.today())} with new FFI delivery.' 
+                # header['HISTORY'] = f'First FFI is {str(os.path.basename(self.file_list[0]))}'
```
falkben, line 459: Should these be commented still?

astrocut/make_cube.py
jaymedina: I'm going to do another PR to add class attributes to the original CubeFactory so we can have 1 analogous CubeFactory for SPOC and TICA, and I'll add an indexing class attribute as well.

astrocut/make_cube.py
jaymedina: This new logic + the everything under the TICA-specific function called `update_cube` is so that the TESS team can add new FFIs onto existing TICA cubes as they get delivered. SPOC deliveries will continue to be only once per month for a single sector, whereas TICA deliveries are 2x a month and will increase to 4x a month starting in September. 

We will make a fresh TICA cube at the beginning of the month so we can get data out faster to users, but we will need to update these cubes as new data comes in throughout the month for a given sector, so that's what this logic is for, and is only specific to TICA.

jaymedina:
Let me know how this is looking and if it's OK to merge @falkben @jwfraustro

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+                    self.template_file = ffi
+                    
+            ffi_data.close()
+
+        self.file_list = file_list[np.argsort(start_times)]
+
+        # Working out the block size and number of blocks needed for writing the cube
+        # without using too much memory
+        slice_size = image_shape[1] * len(self.file_list) * 2 * 4  # in bytes (float32)
+        max_block_size = int((self.max_memory * 1e9)//slice_size)
```
falkben, line 422:
```suggestion
        max_block_size = (self.max_memory * 1e9) // slice_size
```

no need to call `int()` here. realize this was copy / pasted

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+                header = deepcopy(first_file[0].header)
+                header.remove('CHECKSUM', ignore_missing=True)
+
+                # Adding standard keywords
+                header['ORIGIN'] = 'STScI/MAST'
+                header['DATE'] = str(date.today())
+
+                # Adding factory specific keywords
+                for kwd in self.image_header_keywords:
+                    header[kwd] = (first_file[0].header[kwd], first_file[0].header.comments[kwd])
```
falkben, line 448: can you add a comment here about why this is `first_file[0]` instead of `first_file[1]` assuming this is intended?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        else:
+            with fits.open(self.cube_file, mode='update', memmap=True) as cube_hdu:
+                header = cube_hdu[0].header 
+                header['HISTORY'] = f'Updated on {str(date.today())} with new FFI delivery.' 
+                header['HISTORY'] = f'First FFI is {str(os.path.basename(self.file_list[0]))}'
+
+        # Adding the keywords from the last file
+        with fits.open(self.file_list[-1], mode='denywrite', memmap=True) as last_file:
+            for kwd in self.last_file_keywords:
+                header[kwd] = (last_file[0].header[kwd], last_file[0].header.comments[kwd])
```
falkben, line 464: can you add a comment here about why this is last_file[0] instead of last_file[1] assuming this is intended?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+
+    def _build_info_table(self):
+        """ Reading the keywords and setting up the table to hold the image headers (extension 1)
+        from every input file.
+        """
+
+        with fits.open(self.template_file, mode='denywrite', memmap=True) as ffi_data:
+            
+            # The image specific header information will be saved in a table in the second extension
+            secondary_header = ffi_data[0].header
```
falkben, line 478: this is actually the first header, right? can you update variable name and comment if intended?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+
+            # set up the image info table
+            already_there = []
```
falkben, line 481: can we rename this variable `column_names` or something? and then instead of a `continue` we can just simply have a conditional before the `cols.append()`

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+            cols = []
+            for kwd, val, cmt in secondary_header.cards: 
+                if type(val) == str:  
+                    tpe = "S" + str(len(val))  # TODO: Maybe switch to U?
+                elif type(val) == int:
+                    tpe = np.int32
+                else:
+                    tpe = np.float64
+                    
+                length = len(self.file_list)
```
falkben, line 491: can you pull this out of the for loop?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        
+        # Making sure the output directory exists
+        direc, _ = os.path.split(cube_file)
+        if direc and not os.path.exists(direc):
+            os.makedirs(direc)
+
+        # Writing the primary header
+        hdu0 = fits.PrimaryHDU(header=self.primary_header)
+        hdul = fits.HDUList([hdu0])
+        hdul.writeto(cube_file, overwrite=True)
```
falkben, line 523: curious, why the change from:

```python
self.primary_header.tofile(cube_file, overwrite=True)
```

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+
+        # Making the cube header and writing it
+        data = np.zeros((100, 100, 10, 2), dtype=np.float32)
+        hdu = fits.ImageHDU(data)
+        header = hdu.header
+        header["NAXIS4"], header["NAXIS3"], header["NAXIS2"], header["NAXIS1"] = self.cube_shape
+
+        # I think this part increases the buffer size of ext1 and writes 
+        # the cube in there?
```
falkben, line 532: I don't understand this comment. doesn't this just write the header into the cube as bytes?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        # Expanding the file to fit the full data cube
+        # fits requires all blocks to be a multiple of 2880
+        cubesize_in_bytes = ((np.prod(self.cube_shape) * 4 + 2880 - 1) // 2880) * 2880
+        filelen = os.path.getsize(cube_file)
+        
+        with open(cube_file, 'r+b') as CUBE:
+            # What is this .seek?
+            # Looks like it's expanding the buffer size of the file?
+            CUBE.seek(filelen + cubesize_in_bytes - 1)
+            # b for buffer?
```
falkben, line 545: the `b` prefix signifies a "Byte literal". i think this comment and the ones above should be removed. can possibly replace with ...

`# seek to end of file and write null byte`

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        # Loop through files
+        for i, fle in enumerate(self.file_list):
+
+            if verbose:
+                st = time()
+
+            with fits.open(fle, mode='denywrite', memmap=True) as ffi_data:
+
+                # add the image and info to the arrays
+                sub_cube[:, :, i, 0] = ffi_data[0].data[start_row:end_row, :]
```
falkben, line 570: can you add a comment here abou thow this and the lines following are different from the SPOC versions.

Specifically, we're using different indexing and creating only a single `sub_cube`

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+                            # A header keyword ('COMMENT') in TICA returns a keyword 
+                            # value in the form of a _HeaderCommentaryCard instead of a STRING. 
+                            # This breaks the info table because the info table 
+                            # doesn't understand what a FITS header commentary card is. 
+                            # Adding a try/except is a way to catch when these instances happen 
+                            # and turn the keyword value from a _HeaderCommentaryCard to a
+                            # string, which is what it's meant to be in the info table. 
+                            #
+                            # TO-DO: Find a more elegant way to handle these stray 
+                            # _HeaderCommentaryCards.
```
falkben, line 597: a little confused by this.

what doesn't understand a `_headerCommentaryCard`?  aren't these pretty common in astropy tables?  can't we just check for 

```python
isinstance(ffi_data[0].header.get(kwd), fits.header._HeaderCommentaryCards)`
```

and then set to an empty one?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+                            #
+                            # TO-DO: Find a more elegant way to handle these stray 
+                            # _HeaderCommentaryCards. 
+                            
+                            try:
+                                self.info_table[kwd][i] = ffi_data[0].header.get(kwd, nulval)
+                            except ValueError:
+                                kwd_val = ffi_data[0].header.get(kwd)
+                                if isinstance(kwd_val, fits.header._HeaderCommentaryCards):
+                                    self.info_table[kwd][i] = str(kwd_val)
```
falkben, line 604: if we raise a valueerror here and it's not an instance of _headerCOmmentaryCards, we just silently ignore the error.

We should add a `raise` here at the end

```suggestion
                            except ValueError:
                                kwd_val = ffi_data[0].header.get(kwd)
                                if isinstance(kwd_val, fits.header._HeaderCommentaryCards):
                                    self.info_table[kwd][i] = str(kwd_val)
                                else:
                                    raise
```

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        del sub_cube
+    
+    def _update_info_table(self):
+        """ Updating an existing info table with newly created
+        """
+
+        with fits.open(self.template_file, mode='denywrite', memmap=True) as ffi_data:
+            
+            # The image specific header information will be saved in a table in the second extension
+            secondary_header = ffi_data[0].header
```
falkben, line 627: This looks like we're getting it from the first extension

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+            self.cube_append[start_row:end_row, :, :, :] = sub_cube
+
+        if (version_info <= (3, 8)) or (platform == "win32"):
+            cube_hdu.flush()
+
+        del sub_cube
+    
+    def _update_info_table(self):
+        """ Updating an existing info table with newly created
+        """
```
falkben, line 622: Is there some way you can re-use the code in `_build_info_table()` here instead of creating another method with largely the same code?

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        # Adding the comments to the header
+        for kwd in self.info_table.columns:
+            if kwd in ['XTENSION', 'BITPIX', 'NAXIS', 'NAXIS1', 'NAXIS2', 'PCOUNT', 'GCOUNT', 'TFIELDS']:
+                continue  # skipping the keyword already in use
+            table_hdu.header[kwd] = self.info_table[kwd].meta.get("comment", "")
+
+        # Appending to the cube file
+        with fits.open(self.cube_file, mode='update', memmap=True) as cube_hdus:
+            if self.update:
+                cube_hdus.pop(index=2)
```
falkben, line 693: can you add a comment here about why we're doing this?

astrocut/make_cube.py
jaymedina: The TICA FFIs have their science images and science-related keywords in the `0`th extension, whereas SPOC has it in the `1`st

falkben: okay, can you please update the variable name and comment

astrocut/make_cube.py
jaymedina: This part of the code doesn't create an `astropy` table, it creates a FITS table, which doesn't understand what a Header Commentary Card is, so when attempting to feed it the keyword value for `COMMENT` I was getting a `ValueError`. This prompted me to set a conditional in which I turn the comment into a string, because I do want the Header Commentary Card, but I want it in string form so it's legible by the FITS table.

astrocut/make_cube.py
jaymedina: Definitely. I'll add this as a task for when I'm optimizing the code, but for this PR I'll leave it as is because the functionality works OK.

astrocut/make_cube.py
falkben: is there an issue for this?

falkben (commented):
apparently had some pending review comments that I never submitted

astrocut/make_cube.py
```diff
@@ -346,3 +349,500 @@ def make_cube(self, file_list, cube_file="img-cube.fits", sector=None, max_memor
...
+        self.update = True  # we're updating!
+
+        if verbose:
+            startTime = time()
+            
+        # Creating an empty cube that will be appended to the existing cube
+        og_cube = fits.getdata(cube_file, 1)
+        dimensions = list(og_cube.shape)
+        dimensions[2] = len(file_list)
+        self.cube_append = np.zeros(tuple(dimensions))
```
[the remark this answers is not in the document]
jaymedina: Yes: 

```
In [8]: np.zeros(2, 100, 10, 10)
--------------------------------------------------------------
TypeError                    Traceback (most recent call last)
<ipython-input-8-bc4ae52eadb4> in <module>
----> 1 np.zeros(2, 100, 10, 10)

TypeError: zeros() takes from 1 to 3 positional arguments but 4 were given
```

astrocut/make_cube.py
jaymedina: Yes: https://github.com/spacetelescope/astrocut/issues/62

astrocut/make_cube.py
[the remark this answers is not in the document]
falkben: But isn't dimensions a list

falkben (changes requested):

.github/workflows/ci_workflows.yml
```diff
@@ -37,11 +37,6 @@ jobs:
             os: ubuntu-latest
             python: 3.7
             toxenv: py37-test-alldeps-numpy121-cov
-
-          - name: Python 3.8 with all optional dependencies (Windows)
-            os: windows-latest
-            python: 3.8
-            toxenv: py38-test-alldeps
```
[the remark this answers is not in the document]
jaymedina: I'll comment out 3.6 and keep 3.8 deleted, since it will need its own section anyway.

astrocut/tests/test_make_cube.py
```diff
@@ -35,9 +36,9 @@ def test_make_cube(tmpdir):
         ecube[:, :, i, 0] = -plane
         ecube[:, :, i, 1] = plane
         plane += img_sz*img_sz
-
+
```
falkben, line 39:
```suggestion

```

astrocut/tests/test_make_cube.py
```diff
@@ -48,6 +49,56 @@ def test_make_cube(tmpdir):
     hdu.close()
     
 
+def test_update_cube(tmpdir):
```
falkben, line 52: can you rename this?

```suggestion
def test_make_and_update_cube(tmpdir):
```

astrocut/tests/test_make_cube.py
```diff
@@ -48,6 +49,56 @@ def test_make_cube(tmpdir):
...
+    cube_maker = TicaCubeFactory()
+
+    img_sz = 10
+    num_im = 100
+    
+    ffi_files = create_test_ffis(img_sz, num_im, product='tica', dir_name=tmpdir)
+
+    cube_file = path.join(os.getcwd(), "out_dir", "test_update_cube.fits")
+
+    cube_maker.make_cube(ffi_files[0:num_im//2], cube_file, verbose=False)
```
falkben, line 68: since this is our test of the make_cube of TICA, can you add some assertions (like we have for the SPOC cubes above)?

if it makes more sense to split out into a separate test function that's fine too

falkben (changes requested):

astrocut/tests/test_make_cube.py
falkben: you call `make_cube()` with no assertions until after you call `update_cube()`

since these are two separate functions, i would prefer to have some assertions on _just_ the `make_cube()` portion.

jaymedina: Just added some assertions.

Since it's repetitive code, I considered moving them to a separate function and calling that function in for both the `make_cube` and `update_cube` `cube_file`s, but I'm not sure how `pytest` would react to that so it'll need some testing. Let me know if you think it's worth it or if you think it should be left as is. 

As for separating the `make_cube` portion from `update_cube`, I didn't see what there was to gain in doing that, especially since `update_cube` relies on the output of `make_cube`, so it seemed logical to keep them together.

falkben: yes, thank you, this is exactly what i was looking for

falkben (approved):

falkben:
Thanks for all your work on this!

jaymedina:
thanks for the review! merging now

PR #60, opened 2022-04-25
Make cutouts out of TICA cubes

jaymedina:
Implements a class based off the existing CutoutFactory class that can generate TPFs out of TICA cubes.

Note: This is one of the core four projects for MAST Q3 and must be implemented before June.

- [x] Implement TICACutoutFactory
- [x] Verify file structural similarity between TICA and TESS TPFs
- [x] Verify image similarity between TICA and TESS TPFs
- [x] Ensure TPF header keywords should be the same between TICA and TESS
- [x] Ensure TICA header keyword values are unique from TESS and accurate
- [x] Add new header keyword in TPF that lets users know what FFIs were used to make the TPF (TICA or TESS?)
- [x] Unit test(s)


jaymedina:
Note:
TPF header keywords should be the same between TICA and TESS:
Add new header keyword in TPF that lets users know what FFIs were used to make the TPF (TICA or TESS?) 
Add new header keyword showing version of astrocut used to make TPF

jaymedina:
Tests will continue to fail until PR https://github.com/spacetelescope/astrocut/pull/59 gets merged and this branch gets updated with those changes.

astrocut/cube_cut.py
```diff
@@ -98,17 +99,23 @@ def _parse_table_info(self, table_data, verbose=False):
             The cube image header data table.
```
seweissman, line 99: `product` param needs documentation here

astrocut/cube_cut.py
```diff
@@ -98,17 +99,23 @@ def _parse_table_info(self, table_data, verbose=False):
...
         while table_row is None:
             table_row = table_data[data_ind]
-            if table_row["WCSAXES"] != 2:
-                table_row = None
-                data_ind += 1
-                if data_ind == len(table_data):
-                    raise wcs.NoWcsKeywordsFoundError("No FFI rows contain valid WCS keywords.")
+
+            # Making sure we have a row with wcs info.
+            # TODO: Add a check for TICA.
```
seweissman, line 111: Does this TODO need to be done before this can be merged?

astrocut/cube_cut.py
```diff
@@ -440,7 +446,7 @@ def _get_cutout(self, transposed_cube, verbose=True):
         return img_cutout, uncert_cutout, aperture
 
 
-    def _update_primary_header(self, primary_header):
+    def _update_primary_header(self, product, primary_header):
```
seweissman, line 449: `product` parameter needs documentation

astrocut/cube_cut.py
```diff
@@ -457,17 +463,102 @@ def _update_primary_header(self, primary_header):
...
+            primary_header['ASTATE'] = ('N/A', 'archive state F indicates single orbit processi')
+            primary_header['CRMITEN'] = (primary_header['CRM'], 'spacecraft cosmic ray mitigation enabled')
+            primary_header['CRBLKSZ'] = ('N/A', '[exposures] s/c cosmic ray mitigation block siz')
+            primary_header['FFIINDEX'] = (primary_header['CADENCE'], 'number of FFI cadence interval of first FFI')
+            primary_header['DATA_REL'] = ('N/A', 'data release version number')
+
+            date_obs = Time(primary_header['TSTART']+primary_header['BJDREFI'], format='jd').iso
+            date_end = Time(primary_header['TSTOP']+primary_header['BJDREFI'], format='jd').iso
+            primary_header['DATE-OBS'] = (date_obs, 'TSTART as UTC calendar date of first FFI')
+            primary_header['DATE-END'] = (date_end, 'TSTOP as UTC calendar date of last FFI')
```
seweissman, line 507: Are `primary_header['TSTART']` and `primary_header['BJDREFI']` numeric types or strings? I.e. is this a string concat or a numeric addition? Do you need to potentially handle any errors with parsing these dates/times?

astrocut/cube_cut.py
```diff
@@ -457,17 +463,102 @@ def _update_primary_header(self, primary_header):
...
+
+            primary_header['FILEVER'] = ('N/A', 'file format version')
+            primary_header['RADESYS'] = ('N/A', 'reference frame of celestial coordinates')
+            primary_header['SCCONFIG'] = ('N/A', 'spacecraft configuration ID')
+            primary_header['TIMVERSN'] = ('N/A', 'OGIP memo number for file format')
+
+            # Bulk removal with wildcards. Most of these should only live in EXT 1 header.
+            del primary_header['SC_*']  # removes predicted RA, Dec, Roll, etc
```
seweissman, line 515: These header deletions only apply to TICA?

astrocut/cube_cut.py
```diff
@@ -538,6 +629,10 @@ def _add_img_kwds(self, table_header):
         """
 
         for key in self.img_kwds:
+            # We'll skip these TICA-specific image keywords that are single-FFI specific
+            # or just not helpful
+            if (key == 'TIME') | (key == 'EXPTIME') | (key == 'FILTER'):
```
seweissman, line 634:
```suggestion
            if (key == 'TIME') or (key == 'EXPTIME') or (key == 'FILTER'):
```

astrocut/cube_cut.py
seweissman: (There are more places in the code where this has to happen. I'm not going to mark them all.)

astrocut/cube_cut.py
```diff
@@ -457,17 +463,102 @@ def _update_primary_header(self, primary_header):
...
+        timesys_desc = 'Barycentric' if product == 'SPOC' else 'Terrestrial'
+        primary_header['TIMESYS'] = (f'{timesys}', f'time system is {timesys_desc} Dynamical Time ({timesys})')
+
         primary_header['BJDREFI'] = (2457000, 'integer part of BTJD reference date')           
         primary_header['BJDREFF'] = (0.00000000, 'fraction of the day in BTJD reference date')    
         primary_header['TIMEUNIT'] = ('d', 'time unit for TIME, TSTART and TSTOP')
 
+        if product == 'TICA':
+
+            # Adding some missing kwds not in TICA (but in Ames-produced SPOC ffis)
```
seweissman, line 490: Is it worth adding some tests for the header manipulations?

astrocut/cube_cut.py
jaymedina: Correct, I'm deleting only the TICA header keywords that clutter the headers/are not supposed to be in the primary header/are header keywords that describe a single FFI which could be misleading to users, because the cutouts are stacks of FFIs.

astrocut/cube_cut.py
```diff
@@ -40,51 +41,51 @@ def __init__(self):
...
                                'WCS_SIG': [None, "[deg] Error measurement of cutout WCS fit"]}
         
-        self.cutout_lims = np.zeros((2, 2), dtype=int)  # Cutout pixel limits, [[ymin,ymax],[xmin,xmax]]
-        self.center_coord = None  # Central skycoord
+        self.cutout_lims = np.zeros((2, 2), dtype=int) # Cutout pixel limits, [[ymin,ymax],[xmin,xmax]]
+        self.center_coord = None # Central skycoord
         
-        # Extra keywords from the FFI image headers (TESS specific)
+        # Extra keywords from the FFI image headers in SPOC
         self.img_kwds = {"BACKAPP": [None, "background is subtracted"],
```
jeffchamblee, line 53: Would it make sense to read in keywords from a config file? Would it make future maintenance easier?

astrocut/cube_cut.py
jeffchamblee: Would it make sense to read in keywords to be added and deleted from a config file?

astrocut/tests/test_cube_cut.py
```diff
@@ -150,61 +161,64 @@ def test_cutout_extras(tmpdir):
     cube_file = cube_maker.make_cube(ffi_files, path.join(tmpdir, "test_cube.fits"), verbose=False)
 
     # Making the cutout
-    myfactory = CutoutFactory()
     coord = "256.88 6.38"
 
     ###########################
     # Test  _parse_table_info #
```
jeffchamblee, line 167: I recommend breaking up large tests into several smaller tests when possible. That way you can be more selective with @skip & @not_smoke, etc.

astrocut/cube_cut.py
```diff
@@ -31,84 +32,97 @@ class CutoutFactory():
     Class for creating image cutouts.
 
     This class emcompasses all of the cutout functionality.
```
falkben, line 34:
```suggestion
    This class encompasses all of the cutout functionality.  
```

astrocut/cube_cut.py
```diff
@@ -31,84 +32,97 @@ class CutoutFactory():
...
+                        "TMOFST11": [None, "(s) readout delay for camera 1 and ccd 1"],
+                        "VIGNAPP": [None, "vignetting or collimator correction applied"]}
 
         
-    def _parse_table_info(self, table_data, verbose=False):
+    def _parse_table_info(self, product, table_data, verbose=False):
         """
-        Takes the header and one entry from the cube table of image header data,
+        Takes the header and the middle entry from the cube table (EXT 2) of image header data,
         builds a WCS object that encalpsulates the given WCS information,
```
falkben, line 95:
```suggestion
        builds a WCS object that encapsulates the given WCS information,
```

astrocut/cube_cut.py
```diff
@@ -31,84 +32,97 @@ class CutoutFactory():
...
 
     def __init__(self):
         """
-        Initiazation function.
+        Initialization function.
         """
 
-        self.cube_wcs = None  # WCS information from the image cube
-        self.cutout_wcs = None  # WCS information (linear) for the cutout
+        self.cube_wcs = None # WCS information from the image cube
```
falkben, line 47: Having all these whitespace changes are a bit unfortunate. Is there a particular style you are using here? 

`black` style (which will be used in astropy) specifies two spaces between code and comments. If possible, would be nice to revert these changes --- or can just run `black` on it, ideally in a separate PR without any functional changes. 

Same goes for all the changes to the alignment on the `img_kwds`. 

The extra changes make it difficult to review since there's a much larger diff.

astrocut/cube_cut.py
jaymedina: Hi Jeff, yes I agree I think moving things into a config file would help with the troubleshooting. I've created Issue # https://github.com/spacetelescope/astrocut/issues/66 for this so I can tackle it after the initial TICA deployment.

astrocut/cube_cut.py
jaymedina: They're both numeric types, so this was a numeric addition which was then converted to an iso date. I don't anticipate these data types to change, but I've implemented a few checks to make sure these calculations are correct in the latest commit.

jaymedina:
`tica-make-cutouts` is done and ready for final approval.

astrocut/cube_cut.py
```diff
@@ -84,27 +89,34 @@ def __init__(self):
                          "VIGNAPP": [None, "vignetting or collimator correction applied"]}
 
         
-    def _parse_table_info(self, table_data, verbose=False):
+    def _parse_table_info(self, product, table_data, verbose=False):
```
falkben, line 92: we do use this function in tesscut. perhaps we shouldn't be, since it's "internal". or perhaps it should not be an internal function. nevertheless, it's fine as you have it, but it might have been nice to have had product be a keyword argument that defaulted to SPOC instead of being a positional arg so that downstream code wouldn't have to change. I believe you are also taking care of the downstream code, at least in tesscut. Are there any other packages that might use this function? Astroquery?

astrocut/cube_cut.py
```diff
@@ -450,27 +461,102 @@ def _update_primary_header(self, primary_header):
...
+
+            # Adding some missing kwds not in TICA (but in Ames-produced SPOC ffis)
+            primary_header['EXTVER'] = ('1', 'extension version number (not format version)')
+            primary_header['SIMDATA'] = (False, 'file is based on simulated data')
+            primary_header['NEXTEND'] = ('2', 'number of standard extensions')
+            primary_header['TSTART'] = (primary_header['STARTTJD'], 'observation start time in TJD of first FFI')
+            primary_header['TSTOP'] = (primary_header['ENDTJD'], 'observation stop time in TJD of last FFI')
+            primary_header['CAMERA'] = (primary_header['CAMNUM'], 'Camera number')
+            primary_header['CCD'] = (primary_header['CCDNUM'], 'CCD chip number')
+            primary_header['ASTATE'] = ('None', 'archive state F indicates single orbit processi')
```
falkben, line 511: Elsewhere (see https://github.com/spacetelescope/astrocut/blob/1431a64a2ef7eebe377bd8fcfecebd2ee0473f79/astrocut/cube_cut.py#L573) these headers are assigned the actual value `None` instead of the string 'None'. Does it matter which is used?

astrocut/cube_cut.py
```diff
@@ -594,41 +687,49 @@ def _build_tpf(self, cube_fits, img_cube, uncert_cube, cutout_wcs_dict, aperture
...
+        self._update_primary_header(product, primary_hdu.header)
 
         cols = list()
 
         # Adding the cutouts
         tform = str(img_cube[0].size) + "E"
         dims = str(img_cube[0].shape[::-1])
         empty_arr = np.zeros(img_cube.shape)
-
+
```
falkben, line 698:
```suggestion

```

astrocut/cube_cut.py
```diff
@@ -697,6 +797,9 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
             order.  Scalar numbers in ``cutout_size`` are assumed to be in
             units of pixels. `~astropy.units.Quantity` objects must be in pixel or
             angular units.
+        product : str
+            The product type to make the cutouts from.
+            Can either be 'SPOC' or 'TICA'.
```
falkben, line 802: maybe add a comment that default is 'SPOC'?

astrocut/cube_cut.py
```diff
@@ -720,8 +823,13 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
         warnings.filterwarnings("ignore", category=wcs.FITSFixedWarning)
         with fits.open(cube_file, mode='denywrite', memmap=True) as cube:
 
+            # Get the first FFI file, which is used to generate the 
+            # primary header keyword values 
+            self.first_ffi = cube[2].data['FFI_FILE'][0]
+            self.last_ffi = cube[2].data['FFI_FILE'][-1]
```
falkben, line 829: where do these attributes (`first_ffi` and `last_ffi`) get used?

astrocut/tests/test_cube_cut.py
```diff
@@ -49,19 +50,28 @@ def checkcutout(cutfile, pixcrd, world, csize, ecube, eps=1.e-7):
...
     
     # Regression test for PR #6
     assert hdulist[2].data.dtype.type == np.int32
 
     return 
 
 
-def check1(flux, x1, x2, y1, y2, ecube, label, cutfile):
-    """Test one of flux or error"""
+def check1(ffi_type, flux, x1, x2, y1, y2, ecube, label, cutfile):
```
falkben, line 66: `ffi_type` is not accessed in this function

astrocut/cube_cut.py
jaymedina: In cases where a user wants to index one of these keywords, it might be better for them to see a 'None' string as opposed to nothing (the latter would probably make it seem like the HDU is incomplete, so I'll go ahead and change everything to a string.

falkben (changes requested):
looks good, i think just a few small things here and we are ready!

astrocut/cube_cut.py
jaymedina: I made these calls when attempting to add header keywords for these specific files, but the filenames are too long so its been left as a nice-to-have TODO item. I'll remove these calls from this PR since it's now out of scope.

astrocut/cube_cut.py
falkben: i'm fine with not changing this now. just pointing it out that we need to change it in tesscut (also in the s3_support branch, when we switch over to using this)

astrocut/cube_cut.py
jaymedina: No, fixed

astrocut/cube_cut.py
jaymedina: I'm in favor of changing this here to prevent any future bugs elsewhere. I'll work on this next.

astrocut/tests/test_cube_cut.py
jaymedina: The `ffi_type` here is `product` which gets fed from `check_cutout`, so it was working but I've renamed it to `product` to avoid confusion.

falkben: it can be removed as an input from this function.  ffi_type/product is not accessed.

falkben (approved):

PR #70, opened 2022-11-21
hotfix for `cube_cut` checking for valid WCS info

jaymedina:
`_parse_table_info` was calling for a keyword in the cube that does not exist until after the cutout file is made. Updated this WCS validation check accordingly, to work with a header keyword that SPOC and TICA both have, and is always the same value, and implies that other relevant WCS info is also in the header of the FFI. 

Import restructuring from VSCode's `isort`

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -7,7 +7,6 @@ import warnings
 from itertools import product
 from time import time
 
-# Note: Use the astropy function if available, TODO: fix > 4.3 astropy fitting
 import astropy
 import astropy.units as u
 import numpy as np
@@ -19,6 +18,7 @@ from astropy.time import Time
 from . import __version__
 from .exceptions import InputWarning, InvalidQueryError
 
+# Note: Use the astropy function if available, TODO: fix > 4.3 astropy fitting
 if astropy.utils.minversion(astropy, "4.0.2") and (float(astropy.__version__[:3]) < 4.3):
     from astropy.wcs.utils import fit_wcs_from_points
 else:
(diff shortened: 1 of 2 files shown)


dr-rodriguez (dismissed):

astrocut/cube_cut.py
```diff
@@ -4,23 +4,21 @@
 
 import os
 import warnings
-
-from time import time
 from itertools import product
+from time import time
 
+# Note: Use the astropy function if available, TODO: fix > 4.3 astropy fitting
```
falkben, line 10: this comment belongs near line 22

hmlewis-astro (approved):

astrocut/cube_cut.py
```diff
@@ -4,23 +4,21 @@
...
 from astropy.io import fits
 from astropy.time import Time
 
-from . import __version__ 
+from . import __version__
 from .exceptions import InputWarning, InvalidQueryError
 
-# Note: Use the astropy function if available, TODO: fix > 4.3 astropy fitting
-import astropy
 if astropy.utils.minversion(astropy, "4.0.2") and (float(astropy.__version__[:3]) < 4.3):
```
falkben, line 22: Should probably switch to version.parse at some point here...

from packaging import version

Also, I think we can bump the minimum version of astropy. Pretty soon I think we're going to require >= 5.2 for the cloud stuff, so maybe this can just get dropped in a later PR.

falkben (changes requested):

falkben (approved):

astrocut/cube_cut.py
jaymedina: I've made a JIRA ticket to address these two comments in Q2.

PR #76, opened 2023-03-07
Cloud support

falkben:


astrocut/cube_cut.py
```diff
@@ -818,10 +814,21 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
             start_time = time()
 
         warnings.filterwarnings("ignore", category=wcs.FITSFixedWarning)
-        with fits.open(cube_file, mode='denywrite', memmap=True) as cube:
+        fits_options: Dict[str, Any] = {"mode": "denywrite", "lazy_load_hdus": True}
+        if cube_file.startswith("s3://"):
+            fits_options["use_fsspec"] = True
+            # block size should be:
+            # block_size <= m * hdul[1].section.shape[2] * hdul[1].section.shape[3] * 4 bytes
+            fits_options["fsspec_kwargs"] = {"default_block_size": 10_000, "anon": True}
```
falkben, line 822: did we come to a consensus as to what the block size should actually be?

jaymedina: No, there needs to be more analysis on this. I'd like to see total processing time as a function of block size to see what's optimal. Also, wouldnt the optimal block size depend on where the user is accessing our AWS server from? Isnt it that the further away they are from the server Region the slower the request will be?

falkben: I thought we'd decided the most efficient block size would be exactly the "column" of data the user is requesting to cutout?

> Also, wouldn't the optimal block size also depend on where the user is accessing our AWS server from? Isnt it that the further away they are from the server Region the slower the request will be?

Block size efficiency should be irrespective of these things. Being further away, or having a high latency connection, would exacerbate an inefficient block size, though.  But the answer to what the most efficient block size does not vary with these parameters.

falkben: Note: this currently hard codes to ~9.5 KB, which could be too small to be a generalized default, so this probably needs updating.

The default for s3fs is 5 MB (`5 * 2^20`), which is almost certainly too large since the entire cutouts people make can be less than that.

I wonder if this optimization is something we could do in a follow up PR? Either you or I or someone else could handle it?

We can potentially use [`FsspecMonitor()`](https://github.com/barentsen/fsspec-monitor) as verification that we are getting the optimal block size.

See this notebook for usage: https://github.com/fornax-navo/fornax-s3-subsets/blob/main/notebooks/astropy-s3-subsetting-demo.ipynb

There's also some other interesting resources in that repo including some benchmarking code (but not results).

jaymedina (approved):
Looks good, just some minor clarification questions with the tox configurations. I appreciate the test FFIs/cube making process being turned into fixtures. Thanks for the PR!

.github/workflows/ci_workflows.yml
```diff
@@ -16,61 +16,50 @@ jobs:
...
-            
-          - name: Python 3.7 with all optional dependencies (MacOS X)
+            python: "3.10"
+            toxenv: py310-test-alldeps-numpy123-cov
+
+          - name: Python 3.10 with all optional dependencies (MacOS X)
             os: macos-latest
-            python: 3.7
-            toxenv: py37-test-alldeps
+            python: "3.10"
```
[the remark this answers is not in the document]
falkben: because yaml would evaluate this to the number `3.1` without the quotes. and that is not a python release we want to be testing. so we force it to be a string `3.10` which gets us the correct version. it's not necessary for the other versions of python.

.github/workflows/ci_workflows.yml
```diff
@@ -16,61 +16,50 @@ jobs:
...
-      if: ${{ contains(matrix.toxenv,'-cov') }}
-      uses: codecov/codecov-action@v2
-      with:
-        file: ./coverage.xml
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v4
```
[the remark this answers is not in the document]
falkben: latest release of setup-python is v. 4: https://github.com/actions/setup-python (v4.5.0)

.github/workflows/ci_workflows.yml
```diff
@@ -16,61 +16,50 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install language-pack-de tzdata
+      - name: Install Python dependencies
+        run: python -m pip install --upgrade tox codecov
+      - name: Run tests
+        run: tox ${{ matrix.toxargs }} -e ${{ matrix.toxenv }} -- ${{ matrix.toxposargs }}
+      - name: Upload coverage to codecov
+        if: ${{ contains(matrix.toxenv,'-cov') }}
+        uses: codecov/codecov-action@v2
```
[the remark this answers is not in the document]
falkben: there's breaking changes in version 3 that i don't want to deal with in this PR.

i'd be in favor of dropping codecov entirely, actually

astrocut/cube_cut.py
falkben: we spoke about this just recently, but following up here to make it more clear -- I am suggesting we make the s3fs block size configuration dynamic, not fixed, and would be calculated from the specific cutout size requested

PR #81, opened 2023-03-22
split up test and create fixtures for ffi_files and cube_file

falkben:
in an effort to debug a subset of a large test that is currently failing I created a few fixtures and split up the test

i am marking the failing test w/ xfail currently, instead of commenting it out. next step will be trying to address the small cutouts test which is xfailed

### DIFF
--- a/astrocut/tests/test_cube_cut.py
+++ b/astrocut/tests/test_cube_cut.py
@@ -1,5 +1,6 @@
 from os import path
 from pathlib import Path
+from typing import List, Literal
 
 import astropy.units as u
 import numpy as np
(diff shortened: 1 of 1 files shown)


astrocut/tests/test_cube_cut.py
```diff
@@ -231,71 +243,119 @@ def test_cutout_extras(tmp_path, ffi_type):
...
 
-    #############################
-    # Test _get_full_cutout_wcs #
-    #############################
-    cutout_size = [5, 3]
+    assert (xmax - xmin) == 8
+    assert (ymax - ymin) == 15
+
+
+# todo: figure out what this is testing and re-enable
```
jaymedina, line 288: Should remove this `todo` comment

falkben: yup, that's the plan, once i figure out what the problem is with it!

jaymedina: Oh, I thought it was uncommented because it works now

falkben: no, it still does not work. it's marked with xfail

jaymedina: ah yep, see that now. thanks

PR #82, opened 2023-03-22
remove unused parameter in _parse_table_info

falkben:
remove unused parameter in `_parse_table_info`

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -82,19 +82,16 @@ class CutoutFactory():
                          "VIGNAPP": [None, "vignetting or collimator correction applied"]}
 
         
-    def _parse_table_info(self, product, table_data, verbose=False):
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -15,7 +15,7 @@ isolated_build = true
 [testenv]
 
 # Pass through the following environment variables which may be needed for the CI
-passenv = HOME WINDIR LC_ALL LC_CTYPE CC CI TRAVIS
+passenv = HOME,WINDIR,LC_ALL,LC_CTYPE,CC,CI,TRAVIS
```
jaymedina, line 18: this change meant to be part of the pr?

jaymedina (approved):

tox.ini
falkben: yes, otherwise tox won't run at all.

PR #83, opened 2023-04-19
Recalculating block size based on input cutout size

jaymedina:


astrocut/cube_cut.py
```diff
@@ -815,16 +815,23 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
...
             cube_data_prop = "section"
         else:
             fits_options["memmap"] = True
             cube_data_prop = "data"
+
+        # block size should be:
+        # block_size <= m * hdul[1].section.shape[2] * hdul[1].section.shape[3] * 4 bytes
+        if cube_file.startswith("s3://"):
+            with fits.open(cube_file, **fits_options) as cube:
+                block_size = cube[1].section.shape[1] * cube[1].section.shape[2] * cube[1].section.shape[3] * 4
```
falkben, line 832: the m component is dropped here?

falkben, line 832: could you add a test to make sure block size is being set correctly?

astrocut/cube_cut.py
```diff
@@ -815,16 +815,23 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
...
             fits_options["fsspec_kwargs"] = {"default_block_size": 10_000, "anon": True}
             # only use .section for remote data
             cube_data_prop = "section"
         else:
             fits_options["memmap"] = True
             cube_data_prop = "data"
+
+        # block size should be:
+        # block_size <= m * hdul[1].section.shape[2] * hdul[1].section.shape[3] * 4 bytes
+        if cube_file.startswith("s3://"):
```
falkben, line 830: can we move this section inside the conditional on line 819 and it doesn't need to be in it's own location?

astrocut/cube_cut.py
falkben, line 832: i didn't quite think through this entirely and didn't realize we're going to have to read out the header of the cube file each time we make a cutout. i'm wondering how much latency this adds and whether it's worth it versus just taking a hardcoded blocksize approach as we had before. 

can we do some experiments to determine how many requests and how much data is transferred when we set the block size on a per cutout basis (with different cutout sizes and cube sizes) versus just setting it to the same value all the time?  and how much time is added to the cutout by getting the headers here?

falkben (changes requested):
Thanks, this is interesting. Had a few comments/suggestions.

Probably good to have a description as well in the pull request

astrocut/cube_cut.py
jaymedina: The m component is `cube[1].section.shape[1]`

jaymedina: Yeah I can do that. Originally I thought the intention was to hard-code a default block size for all requests. I thought that was why we were trying to determine the "optimal block size" , so we can hard-code a block size that is optimal in terms of request time while also limiting the # of requests to the server.

falkben: hm, i thought it would be dependent on the size of the cutout for some reason. 

anyways, this results in a blocksize much larger than what we'd had before (~6 thousand times larger) for the cube `s3://stpubdata/tess/public/mast/tess-s0038-2-2-cube.fits`

I'm seeing: 10,000 vs. 63,311,040

so if this is correct, we should see _MUCH_ fewer requests. And hopefully also about the same amount of bytes downloaded. Are you seeing this?

falkben: note, we already do read some table info from the cube in `_parse_table_info`. It would be convenient if we could somehow combine some of this metadata extraction into a single remote read.

jaymedina: This happens after the cube file is opened with `**fits_options` tho.

astrocut/cube_cut.py
```diff
@@ -815,16 +815,22 @@ def cube_cut(self, cube_file, coordinates, cutout_size,
...
-            # block size should be:
-            # block_size <= m * hdul[1].section.shape[2] * hdul[1].section.shape[3] * 4 bytes
             fits_options["fsspec_kwargs"] = {"default_block_size": 10_000, "anon": True}
             # only use .section for remote data
             cube_data_prop = "section"
+
+            # block size should be:
+            # block_size <= m * hdul[1].section.shape[2] * hdul[1].section.shape[3] * 4 bytes
+            with fits.open(cube_file, **fits_options) as cube:
+                block_size = cube[1].section.shape[1] * cube[1].section.shape[2] * cube[1].section.shape[3] * 4
```
falkben, line 828: just confirming this should always be `1` here? it doesn't matter if it's TICA or SPOC?

```python
                block_size = cube[1].section.shape[1] * cube[1].section.shape[2] * cube[1].section.shape[3] * 4
```

jaymedina: Correct, I designed `TicaCubeFactory` so that the TICA cubes are structurally identical to the SPOC cubes. EXT1 is always the SCI extension.

astrocut/cube_cut.py
falkben: yes, i'm aware. i was wondering if we should refactor. i was thinking that maybe we open the file once to read all the metadata that we need.

and then we open it once more to actually make the cutout, with the precise block size.

maybe it's not worth worrying about. i am concerned about all the extra access to these remote files being very slow -- so if there's opportunities to combine those together i think we could benefit.

jaymedina: One thing worth mentioning is that making the block size a function of the users input cutout size (as opposed to using a fixed one) adds a few steps extra steps, especially if the cutout size is fed in any unit other than pixels. The script needs to first convert the cutout size to pixels in order to calculate the `block_size` correctly. I added a new method called `_convert_cutout_size` to handle this, and I'll do some analysis tonight on how much time all of these steps add to the request(s). I'll update this thread with the results.

jaymedina: OK. FWIW, these are the times I've been seeing with and without computing a custom block size based on input cutout size (when `custom_block_size` is turned off, the block size is just the original 10,000 bytes: 
image

falkben: That indicates to me that the custom block size computation is likely incorrect.

Can you print out the computed block sizes?

How about enabling Geert's context manager that reports on bytes/requests.

falkben:
I think we can close, based on our conversations on this?

PR #84, opened 2023-06-13
cutout in threadpool

falkben:
adds `threads` to the CutoutFactory.__init__() method which controls whether the threadpool is used.

Using this benchmark code:

```python
"""Benchmark threading performance"""

import warnings
from tempfile import TemporaryDirectory

from astropy.coordinates import SkyCoord
from astropy.io.fits.verify import VerifyWarning

from astrocut.cube_cut import CutoutFactory

warnings.filterwarnings("ignore", category=VerifyWarning)

coord = SkyCoord(217.42893801, -62.67949189, unit="deg", frame="icrs")
cube_file = "s3://stpubdata/tess/public/mast/tess-s0038-2-2-cube.fits"
tmp_path = TemporaryDirectory()


def cutout(threads=None):
    cutout_file = CutoutFactory(threads=threads).cube_cut(cube_file, coord, 31, output_path=str(tmp_path))
    return cutout_file


if __name__ == "__main__":
    import timeit

    # todo: np.array_equal to make sure we're getting the same thing
    # np.array_equal(ff, dd)

    bmark_val = min(timeit.repeat("cutout()", globals=locals(), repeat=5, number=1))
    print(f"cutout threads None: {bmark_val:2.2f}s")

    bmark_val = min(timeit.repeat("cutout(threads=0)", globals=locals(), repeat=5, number=1))
    print(f"cutout no threads: {bmark_val:2.2f}s")

    bmark_val = min(timeit.repeat("cutout(threads=4)", globals=locals(), repeat=5, number=1))
    print(f"cutout threads 4: {bmark_val:2.2f}s")
```

I observe:

cutout threads None: 2.74s
cutout no threads: 3.85s
cutout threads 4: 2.75s

Since this is network IO, there is some variability. With a 31x31 cutout, it's usually at least 1 second faster. This makes a large impact for tesscut.

Potentially in place of https://github.com/astropy/astropy/pull/14932

todo: some validation of cutout contents

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -40,11 +40,6 @@ class CutoutFactory():
     def __init__(self):
         """
         Initialization function.
-
-        Parameters
-        ----------
-        table_data : `~astropy.io.fits.fitsrec.FITS_rec`
-            The cube image header data table.
         """
 
         self.cube_wcs = None  # WCS information from the image cube
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -4,9 +4,10 @@
 
 import os
 import warnings
+from concurrent.futures import ThreadPoolExecutor
 from itertools import product
 from time import time
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 import astropy.units as u
 import numpy as np
@@ -36,10 +37,22 @@ class CutoutFactory():
     Future versions will include more generalized cutout functionality.
     """
 
-    def __init__(self):
+    def __init__(self, threads: Union[int, None] = None):
         """
         Initialization function.
+
+        Parameters
+        ----------
+        table_data : `~astropy.io.fits.fitsrec.FITS_rec`
+            The cube image header data table.
+
+        threads : int or `None`
+            Number of threads to use when making cutouts.
+            <=1 disables the threadpool, > 1 sets the number of threads,
+            None (default) uses Python's default threads
         """
+
+        self.threads = threads
 
         self.cube_wcs = None  # WCS information from the image cube
         self.cutout_wcs = None  # WCS information (linear) for the cutout
@@ -424,8 +437,14 @@ class CutoutFactory():
             ymax = ymax_cube       
         
         # Doing the cutout
-        cutout = transposed_cube[xmin:xmax, ymin:ymax, :, :]
-    
+        if self.threads is None or self.threads > 1:
+            with ThreadPoolExecutor(max_workers=self.threads) as pool:
+                cutouts = list(pool.map(lambda x: transposed_cube[x, ymin:ymax, :, :], range(xmin, xmax)))
+            # stack the list of cutouts
+            cutout = np.stack(cutouts)
+        else:
+            cutout = transposed_cube[xmin:xmax, ymin:ymax, :, :]
+
         img_cutout = cutout[:, :, :, 0].transpose((2, 0, 1))
         uncert_cutout = cutout[:, :, :, 1].transpose((2, 0, 1))
     
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -37,7 +37,7 @@ class CutoutFactory():
     Future versions will include more generalized cutout functionality.
     """
 
-    def __init__(self, threads: Union[int, None] = None):
+    def __init__(self, threads: Union[int, None] = 1):
         """
         Initialization function.
 
@@ -46,10 +46,10 @@ class CutoutFactory():
         table_data : `~astropy.io.fits.fitsrec.FITS_rec`
             The cube image header data table.
 
-        threads : int or `None`
+        threads : int or `None`, default=1
             Number of threads to use when making cutouts.
-            <=1 disables the threadpool, > 1 sets the number of threads,
-            None (default) uses Python's default threads
+            <=1 disables the threadpool, > 1 sets to the specified number of threads,
+            None uses `ThreadPoolExecutor`'s default: cpu_count + 4, limited to max of 32
         """
 
         self.threads = threads
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -7,7 +7,7 @@ import warnings
 from concurrent.futures import ThreadPoolExecutor
 from itertools import product
 from time import time
-from typing import Any, Dict, Union
+from typing import Any, Dict, Literal, Union
 
 import astropy.units as u
 import numpy as np
@@ -37,7 +37,7 @@ class CutoutFactory():
     Future versions will include more generalized cutout functionality.
     """
 
-    def __init__(self, threads: Union[int, None] = 1):
+    def __init__(self):
         """
         Initialization function.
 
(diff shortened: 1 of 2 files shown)


astrocut/cube_cut.py
```diff
@@ -46,10 +46,10 @@ def __init__(self, threads: Union[int, None] = None):
         table_data : `~astropy.io.fits.fitsrec.FITS_rec`
             The cube image header data table.
 
-        threads : int or `None`
+        threads : int or `None`, default=1
             Number of threads to use when making cutouts.
-            <=1 disables the threadpool, > 1 sets the number of threads,
-            None (default) uses Python's default threads
+            <=1 disables the threadpool, > 1 sets to the specified number of threads,
```
scfleming, line 51: Since the default is to not use the thread pool, can we schedule a backlog item to make it easy to find within the Astrocut documentation how a user enables this and why they may want to do so? @jaymedina

falkben:
I did some basic memory profiling on my mac using [memray](https://bloomberg.github.io/memray/) to confirm that a threaded implementation doesn't increase memory consumption too much. Since the threads make multiple requests concurrently into s3, I would have expect some increase, but it shouldn't be too much.

Unfortunately, I cannot upload the report (html) here as github doesn't support it. But here's the plot for each:

|no threads|4 threads|16 threads|
|-|-|-|
| newplot(2) | newplot(3) | newplot(4)  |

memray warns that results on macos are not as precise as those on linux. i may try to do this same test again on a linux VM

falkben:
memray results on a linux VM:

|no threads|4 threads|16 threads|
|-|-|-|
| newplot(5) | newplot(6)  | newplot(7)  |

This shows the small increase in memory i would expect for increasing threads.

astrocut/cube_cut.py
```diff
@@ -424,8 +437,14 @@ def _get_cutout(self, transposed_cube, verbose=True):
             ymax = ymax_cube       
         
         # Doing the cutout
-        cutout = transposed_cube[xmin:xmax, ymin:ymax, :, :]
-    
+        if self.threads is None or self.threads > 1:
+            with ThreadPoolExecutor(max_workers=self.threads) as pool:
+                cutouts = list(pool.map(lambda x: transposed_cube[x, ymin:ymax, :, :], range(xmin, xmax)))
+            # stack the list of cutouts
+            cutout = np.stack(cutouts)
```
seweissman, line 444: I think it would help to have a few comments here on how we are dividing up the cut-out across the threads. It seems like we are slicing it by the x dimension. Does that mean we don't get much benefit from the threading if the number of columns in the cutout is small?

falkben: I can add some comments here. This PR essentially is in place of one to astropy where I had more details of what the issue was. https://github.com/astropy/astropy/pull/14932

astrocut/tests/test_cube_cut.py
```diff
@@ -527,3 +527,30 @@ def test_s3_cube_cut(tmp_path: Path):
...
+
+    # 1 thread condition (no threading) used as verification for the different thread conditions
+    cutf_no_threads = CutoutFactory(threads=1)
+    cutout_no_threads = cutf_no_threads.cube_cut(
+        cube_file, coordinates=coord, output_path=tmpdir, verbose=False, cutout_size=cutout_size
+    )
+    data_no_threads = fits.getdata(cutout_no_threads)
+
+    # when threads=None, # threads is system dependent: cpu_count + 4, limited to max of 32
+    # https://github.com/python/cpython/blob/89ec0e952965b6a1be40e26c3ddc4131599e5ee9/Lib/concurrent/futures/thread.py#L137-L147
```
jaymedina, line 547: Is there a better way to reference this behavior instead of pointing to specific lines of code? In case the developers of this package end up modifying their `thread.py` script a bit more. Might be best to just point to [this documentation](https://docs.python.org/3/library/concurrent.futures.html) which has a history of the previous versions and how `make_workers` was calculated before, and note that we are using version 3.8 for Astrocut.

astrocut/tests/test_cube_cut.py
```diff
@@ -527,3 +527,30 @@ def test_s3_cube_cut(tmp_path: Path):
...
+    for threads in (4, None):
+        cutf_threads = CutoutFactory(threads=threads)
+        cutout_threads = cutf_threads.cube_cut(
+            cube_file, coordinates=coord, output_path=tmpdir, verbose=False, cutout_size=cutout_size
+        )
+        data_threads = fits.getdata(cutout_threads)
+
+        for ext_name in ("FLUX", "FLUX_ERR"):
+            assert np.array_equal(data_no_threads[ext_name], data_threads[ext_name])
```
jaymedina, line 556: Looks good! Do we want to compare the time it took to get `cutout_no_threads` and `cutout_threads` to ensure that `cutout_threads` got returned quicker?

jaymedina (changes requested):

astrocut/tests/test_cube_cut.py
falkben: That link was a permalink, so modifications to thread.py wouldn't change, but I can use this instead: https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor

astrocut/cube_cut.py
```diff
@@ -36,11 +37,23 @@ class CutoutFactory():
...
+
+        Parameters
+        ----------
+        table_data : `~astropy.io.fits.fitsrec.FITS_rec`
+            The cube image header data table.
+
+        threads : int or `None`, default=1
+            Number of threads to use when making cutouts.
+            <=1 disables the threadpool, > 1 sets to the specified number of threads,
+            None uses `ThreadPoolExecutor`'s default: cpu_count + 4, limited to max of 32
```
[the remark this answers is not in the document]
falkben: >  it sounds like this branch is using concurrent version 3.8

do you mean 3.8 and above?  or do you mean 3.8 specifically?

I don't think there's anything specific about 3.8 here.

> I find it a bit confusing that an input of None and an input of <=1 (for example 0) would not configure the number of threads the same way. 

Since the interface to `concurrent.futures` sets max_threads to be automatic when `max_workers=None`, I was sort of thinking it made sense to make our parameter match that. But I could change the auto behavior to only work for the string "auto" instead if that makes more sense. If I switch to "auto" then I guess `None` could instead be no multithreading? Or would you prefer I remove None from the type annotation entirely?

astrocut/cube_cut.py
falkben: I would prefer to have the documentation commits in a separate PR where we can review them separately.

astrocut/tests/test_cube_cut.py
jaymedina: Yeah, I'd point to that `ThreadPoolExecutor` subpage instead.

astrocut/cube_cut.py
falkben: I added several lines of comments that should explain things and link back to the function in astropy that we're trying to optimize

astrocut/cube_cut.py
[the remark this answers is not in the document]
falkben: we only support 3.8 and above so I think this should be fine. the param uses "auto" now.

astrocut/cube_cut.py
seweissman: Thanks. I took a look at that PR and I may be more confused now. I think I just don't understand the underlying data structure...

astrocut/cube_cut.py
jaymedina: Sure thing @falkben , I've created a Jira ticket for this that I've moved to this sprint, so I'll tackle it this week and tag you when ready.

astrocut/cube_cut.py
```diff
@@ -424,8 +431,21 @@ def _get_cutout(self, transposed_cube, verbose=True):
             ymax = ymax_cube       
         
         # Doing the cutout
-        cutout = transposed_cube[xmin:xmax, ymin:ymax, :, :]
-    
+        if threads == "auto" or threads > 1:
```
jaymedina, line 434: One more thing: The multithread logic only works when the input cube is on the cloud, correct? What happens when a locally-stored cube gets fed, and `threads` is set to >1? Perhaps we should add an extra conditional here that ensures`transposed_cube` contains `stpubdata` in the path string, or something like that.

jaymedina (changes requested):

astrocut/cube_cut.py
falkben: > What happens when a locally-stored cube gets fed, and threads is set to >1? 

threads is reassigned to 1 if it doesn't start with s3

> Perhaps we should add an extra conditional here that ensures the input cube contains stpubdata in the path string, or something like that.

I don't think I understand, why would we do that?

jaymedina: > threads is reassigned to 1 if the cube file doesn't start with s3

Ah okay, I see that piece of the code now. Nvm, thanks!

jaymedina (approved):
Looks good !

PR #85, opened 2023-07-14
Error handling for `CubeFactory` and `TicaCubeFactory`

jaymedina:
### Summary 

Currently the errors that get returned when SPOC FFIs are fed into `TicaCubeFactory` and vice versa are vague and not helpful to the user. This PR updates the error handling by raising an error when the wrong type of FFI is fed into either `CubeFactory` or `TicaCubeFactory` to make TESS cubes.

----

### Task List

- [x] Update `CubeFactory` error handling 
- [x] Update `TicaCubeFactory` error handling
- [x] Add unit test(s)

----

### Screenshots (as of July 14, 2023)

The following two screenshots show the error that gets thrown when feeding the wrong type of FFI to `CubeFactory` and `TicaCubeFactory`:

image

image

### DIFF
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -19,8 +19,7 @@ if (version_info >= (3, 8)) and (platform != "win32"):
     from mmap import MADV_SEQUENTIAL
 
 __all__ = ['CubeFactory', 'TicaCubeFactory']
-ERROR_MSG = "One or more incorrect file types were input. Please input TICA FFI files when using\
-                   ``TicaCubeFactory``, and SPOC FFI files when using ``CubeFactory``."
+ERROR_MSG = "One or more incorrect file types were input."
 
 class CubeFactory():
     """
--- a/astrocut/make_cube.py
+++ b/astrocut/make_cube.py
@@ -19,7 +19,9 @@ if (version_info >= (3, 8)) and (platform != "win32"):
     from mmap import MADV_SEQUENTIAL
 
 __all__ = ['CubeFactory', 'TicaCubeFactory']
-ERROR_MSG = "One or more incorrect file types were input."
+ERROR_MSG = "One or more incorrect file types were input. Please input TICA FFI files when using\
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -17,6 +17,7 @@ python_requires = >=3.6
 setup_requires = setuptools_scm
 install_requires =
     astropy>=5.2 # astropy with s3fs support
+    astroquery>=0.4.6 # for testing
```
falkben, line 20: if this is only needed for testing, can it be moved into the test extra section below?

astrocut/make_cube.py
```diff
@@ -85,7 +85,10 @@ def _configure_cube(self, file_list, **extra_keywords):
 
         # Working out the block size and number of blocks needed for writing the cube
         # without using too much memory
-        slice_size = image_shape[1] * len(self.file_list) * 2 * 4  # in bytes (float32)
+        try:
+            slice_size = image_shape[1] * len(self.file_list) * 2 * 4  # in bytes (float32)
+        except IndexError:
+            raise ValueError("One or more TICA FFIs were input. Please use ``TicaCubeFactory`` to process TICA FFIs.")
```
falkben, line 91: I wonder if we can be so certain this is the error? couldn't an indexerror be caused by inputting other types of files? Perhaps the error message could be more general, and suggest that maybe the input is of the wrong ffi type?

setup.cfg
jaymedina: Sure thing! I'll address these comments when I fix the CI. Thanks!

astrocut/make_cube.py
jaymedina: It depends on the structure of the FITS file, and whether or not the input file is structured the same as SPOC (for `CubeFactory`) and TICA (for `TicaCubeFactory`). But yes, in theory, this could break due to a non-TESS file. I've updated the error message accordingly.

hmlewis-astro (approved):

PR #86, opened 2023-07-18
Documenting multithreading enhancement

jaymedina:
This PR updates the Astrocut documentation to reflect the work done in PR https://github.com/spacetelescope/astrocut/pull/84

These are the new changes:

image

### DIFF
--- a/docs/astrocut/index.rst
+++ b/docs/astrocut/index.rst
@@ -261,6 +261,31 @@ which will allow you the option to implement multiprocessing to improve cutout g
 speed. To use multiprocessing, set the ``threads`` argument in ``cube_cut`` to be greater than 1, or set ``threads``
 to ``"auto"``, which will set the number of threads based on the CPU count of your machine.
 
+.. code-block:: python
+
+                >>> from astrocut import CutoutFactory
+                >>> from astropy.coordinates import SkyCoord
+
+                >>> my_cutter = CutoutFactory()
+                >>> coord = SkyCoord(217.42893801, -62.67949189, unit="deg", frame="icrs")
+                >>> cutout_size = 30
+                >>> cube_file = "s3://stpubdata/tess/public/mast/tess-s0038-2-2-cube.fits"
+
+                >>> cut_factory.cube_cut(cube_file, coordinates=coord, cutout_size=cutout_size, verbose=True, threads="auto") #doctest: +SKIP
+                Using WCS from row 1852 out of 3705
+                Cutout center coordinate: 217.42893801,-62.67949189
+                xmin,xmax: [1572 1602]
+                ymin,ymax: [852 882]
+                Image cutout cube shape: (3705, 30, 30)
+                Uncertainty cutout cube shape: (3705, 30, 30)
+                Maximum distance between approximate and true location: 3.6009402965268847e-05 deg
+                Error in approximate WCS (sigma): 0.0003207242331953156
+                Target pixel file: ./tess-s0038-2-2_217.428938_-62.679492_30x30_astrocut.fits
+
+                WARNING: VerifyWarning: Card is too long, comment will be truncated. [astropy.io.fits.card]
+
+                Write time: 0.54 sec
+                Total time: 4.3 sec
   
 Additional Cutout Processing
 ============================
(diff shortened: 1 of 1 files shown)


docs/astrocut/index.rst
```diff
@@ -219,7 +219,7 @@ Making cutout target pixel files
 --------------------------------
 
 To make a cutout, you must already have an image cube to cut out from.
-Assuming that that step has been completed, you simply give the central
+Assuming that you have a TESS cube file stored locally, you can simply give the central
```
falkben, line 222: I know not your change, but can we remove all instances of the word "simply" from our docs?

docs/astrocut/index.rst
```diff
@@ -256,7 +256,60 @@ The cutout target pixel file format is decribed `here <file_formats.html#target-
...
+
+                >>> from astrocut import CutoutFactory
+                >>> from astropy.coordinates import SkyCoord
+
+                >>> my_cutter = CutoutFactory()
+                >>> coord = SkyCoord(217.42893801, -62.67949189, unit="deg", frame="icrs")
+                >>> cutout_size = 30
+                >>> cube_file = "s3://stpubdata/tess/public/mast/tess-s0038-2-2-cube.fits"
+
+                >>> cut_factory.cube_cut(cube_file, coordinates=coord, cutout_size=cutout_size, verbose=True, threads="auto") #doctest: +SKIP
```
falkben, line 275: Can we put some line breaks here so the `threads` option doesn't scroll off the side and is more easily visible?

falkben (changes requested):
This looks really good, thanks for adding this.

scfleming:
"To use multiprocessing, set the `threads` ... argument to an integer greater than 1".  I assume this is actully the number of threads you want to use, as written it makes it sounds like any interger greater than 1 just turns on the feature.  Perhaps we can say "To use multiprocessing, set the `threads` .... argument to the number of threads you want to use.  Setting it to a value of 1 disables the use of threads(? is that right?).  Setting it to `auto`..."

scfleming:
Towards the end it says "By default multithreading is disabled by default, but you can also disable it by setting `threads` <= 1".  At first I misread this and thought it was backwards but noticed it says "disable it" and not "enable it".  If the default is to not enable it, we could just indicate how to enable it (by setting threads > 1) and avoid discussing alternative ways of disabling it to avoid confusing users?  The behavior itself is good (that's naturally what someone should expect this to do if given a # of threads of 1 or less) but I don't feel like it's instructive for users to know that in the documentation.

jaymedina:
> If the default is to not enable it, we could just indicate how to enable it (by setting threads > 1) and avoid discussing alternative ways of disabling it to avoid confusing users? The behavior itself is good (that's naturally what someone should expect this to do if given a # of threads of 1 or less) but I don't feel like it's instructive for users to know that in the documentation.

Yep, that's fair, I've removed the comment about how to disable multiprocessing, and just mention that multiprocessing is disabled by default, and updated the second example (without multiprocessing) accordingly.

docs/astrocut/index.rst
```diff
@@ -256,7 +256,62 @@ The cutout target pixel file format is decribed `here <file_formats.html#target-
                 1  PIXELS        1 BinTableHDU    222   144R x 12C   [D, E, J, 25J, 25E, 25E, 25E, 25E, J, E, E, 38A]   
                 2  APERTURE      1 ImageHDU        45   (5, 5)   float64  
 
-  
+You can also create cutout target pixel files out of TESS cube files stored on MAST's AWS open data bucket.
```
falkben, line 259: Should we add a section header above this?

falkben (changes requested):

docs/astrocut/index.rst
jaymedina: I made it a subsection so that it still falls under the `TESS Full-Frame Image Cutouts` section. And the `Multithreading` subsection is just below it. Here's what it looks like now: 

image

falkben: Can we call this multithreading instead? since multiprocessing and multithreading are very different things, people might get confused if we use the other term.

hmlewis-astro (changes requested):
Just a few places to go through and change multiprocessing -> multithreading, but otherwise, looks good!

docs/astrocut/index.rst
jaymedina: Sure thing

hmlewis-astro (approved):

falkben (approved):

PR #87, opened 2023-07-20
removing error array dimension from TicaCubeFactory

hmlewis-astro:
Addressing ticket ASB-22318. Created a test cube using 500 FFIs from TICA Sector 67, cam1-ccd1. Created cube has expected dimensions, see output from `fitsinfo` below. Cube created without error array is ~1/2 the size of a cube created with the error array (8.68 vs 17.35 GB).
 ```
Filename: /home/halewis/hlsp/tica/tica_cubes/s0067/tica-s0067-1-1-cube.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU     198   ()      
  1                1 ImageHDU         9   (1, 500, 2136, 2078)   float32   
  2                1 BinTableHDU    598   500R x 198C   [D, J, J, J, J, D, J, J, 2A, D, D, D, D, D, D, D, D, D, D, D, D, J, J, J, J, J, J, J, D, J, D, D, D, D, D, D, D, D, 3A, J, J, 16A, D, D, D, D, 9A, 5A, D, 15A, 4A, 4A, D, D, D, D, D, D, D, D, 12A, 12A, D, D, D, D, J, J, J, J, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, D, J, J, J, 16A, 10A, 49A, 64A]
```

Issue #62, opened 2022-08-05
Use class inheritance to merge `TicaCubeFactory` with `CubeFactory`


jaymedina:
Suggestion from @falkben RE: Pull Request #59 

_"Can we use class inheritance here? It looks like most of the changes would just be in the class attributes. If we make those changes, I think we can make these changes much simpler and should be easier to maintain."_

PR #88, opened 2023-07-26
Adapting `CutoutFactory` to account for error-less TICA Cubes

jaymedina:
### Summary

This PR updates `CutoutFactory` to account for the removed error array in `TicaCubeFactory` (work that was done in  #87). It involves turning the `product` kwarg from `cube_cut()` into a class attribute so that it doesn't have to be fed into every method as an argument before being called. It also involves using this new `self.product` class attribute to make conditionals that make the code dynamic and changes the logic depending on whether the product for the cutouts is `"TICA"` or `"SPOC"`. Unit test changes are also made where appropriate.

----

### Task List

- [x] Account for removed error array in `CutoutFactory`
- [x] Update `cube_cut` unit tests
- [x] Rebase onto #87

----

### Screenshots

### DIFF
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -448,7 +448,8 @@ class CutoutFactory():
             cutout = transposed_cube[xmin:xmax, ymin:ymax, :, :]
 
         img_cutout = cutout[:, :, :, 0].transpose((2, 0, 1))
-        uncert_cutout = cutout[:, :, :, 1].transpose((2, 0, 1))
+        if self.product == "SPOC":
+            uncert_cutout = cutout[:, :, :, 1].transpose((2, 0, 1))
     
         # Making the aperture array
         aperture = np.ones((xmax-xmin, ymax-ymin), dtype=np.int32)
@@ -456,14 +457,19 @@ class CutoutFactory():
         # Adding padding to the cutouts so that it's the expected size
         if padding.any():  # only do if we need to pad
             img_cutout = np.pad(img_cutout, padding, 'constant', constant_values=np.nan)
-            uncert_cutout = np.pad(uncert_cutout, padding, 'constant', constant_values=np.nan)
+            if self.product == "SPOC":
+                uncert_cutout = np.pad(uncert_cutout, padding, 'constant', constant_values=np.nan)
             aperture = np.pad(aperture, padding[1:], 'constant', constant_values=0)
 
         if verbose:
             print("Image cutout cube shape: {}".format(img_cutout.shape))
-            print("Uncertainty cutout cube shape: {}".format(uncert_cutout.shape))
+            if self.product == "SPOC":
+                print("Uncertainty cutout cube shape: {}".format(uncert_cutout.shape))
     
-        return img_cutout, uncert_cutout, aperture
+        if self.product == "SPOC":
+            return img_cutout, uncert_cutout, aperture
+        else:
+            return img_cutout, aperture
 
 
     def _update_primary_header(self, primary_header):
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -681,7 +681,8 @@ class CutoutFactory():
         img_cube : `numpy.array`
             The untransposed image cutout array
         uncert_cube : `numpy.array`
-            The untransposed uncertainty cutout array
+            The untransposed uncertainty cutout array.
+            This value is set to `None` by default for TICA cutouts.
         cutout_wcs_dict : dict
             Dictionary of wcs keyword/value pairs to be added to each array 
             column in the cutout table header.
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -702,8 +702,6 @@ class CutoutFactory():
         # The primary hdu is just the main header, which is the same
         # as the one on the cube file
         primary_hdu = cube_fits[0]
-        print(cube_fits)
-        print(primary_hdu.header)
         self._update_primary_header(primary_hdu.header)
 
         cols = list()
@@ -849,6 +847,7 @@ class CutoutFactory():
         if verbose:
             start_time = time()
 
+        # Declare the product type being used to make the cutouts
         self.product = product
 
         warnings.filterwarnings("ignore", category=wcs.FITSFixedWarning)
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -729,10 +729,10 @@ class CutoutFactory():
 
         # Adding flux and flux_err (data we actually have!)
         pixel_unit = 'e-/s' if self.product == 'SPOC' else 'e-'
+        flux_err_array = uncert_cube if self.product == 'SPOC' else empty_arr
         cols.append(fits.Column(name='FLUX', format=tform, dim=dims, unit=pixel_unit, disp='E14.7', array=img_cube))
-        if uncert_cube is not None:
-            cols.append(fits.Column(name='FLUX_ERR', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
-                                    array=uncert_cube)) 
+        cols.append(fits.Column(name='FLUX_ERR', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
+                                array=flux_err_array)) 
    
         # Adding the background info (zeros b.c we don't have this info)
         cols.append(fits.Column(name='FLUX_BKG', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -720,7 +720,8 @@ class CutoutFactory():
                                     array=cube_fits[2].columns['BARYCORR'].array))
 
         # Adding CADENCENO as zeros for SPOC b/c we don't have this info
-        cadence_array = empty_arr[:, 0, 0] if self.product == 'SPOC' else cube_fits[2].columns['CADENCE'].array
+        cadence_array = np.zeros(img_cube.shape)[:, 0, 0] if self.product == 'SPOC'\
+                        else cube_fits[2].columns['CADENCE'].array
         cols.append(fits.Column(name='CADENCENO', format='J', disp='I10', array=cadence_array))
 
         # Adding counts (-1 b/c we don't have data)
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -672,7 +672,7 @@ class CutoutFactory():
                         hdu.header[kwd] = (primary_header[kwd], primary_header.comments[kwd])
             
 
-    def _build_tpf(self, cube_fits, img_cube, uncert_cube, cutout_wcs_dict, aperture):
+    def _build_tpf(self, cube_fits, img_cube, cutout_wcs_dict, aperture, uncert_cube=None):
         """
         Building the cutout target pixel file (TPF) and formatting it to match TESS pipeline TPFs.
 
@@ -702,7 +702,9 @@ class CutoutFactory():
         # The primary hdu is just the main header, which is the same
         # as the one on the cube file
         primary_hdu = cube_fits[0]
-        self._update_primary_header(self.product, primary_hdu.header)
+        print(cube_fits)
+        print(primary_hdu.header)
+        self._update_primary_header(primary_hdu.header)
 
         cols = list()
 
@@ -731,8 +733,9 @@ class CutoutFactory():
         # Adding flux and flux_err (data we actually have!)
         pixel_unit = 'e-/s' if self.product == 'SPOC' else 'e-'
         cols.append(fits.Column(name='FLUX', format=tform, dim=dims, unit=pixel_unit, disp='E14.7', array=img_cube))
-        cols.append(fits.Column(name='FLUX_ERR', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
-                                array=uncert_cube)) 
+        if uncert_cube:
+            cols.append(fits.Column(name='FLUX_ERR', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
+                                    array=uncert_cube)) 
    
         # Adding the background info (zeros b.c we don't have this info)
         cols.append(fits.Column(name='FLUX_BKG', format=tform, dim=dims, unit=pixel_unit, disp='E14.7',
@@ -898,9 +901,14 @@ class CutoutFactory():
                 print("ymin,ymax: {}".format(self.cutout_lims[0]))
 
             # Make the cutout
-            img_cutout, uncert_cutout, aperture = self._get_cutout(
+            results = self._get_cutout(
                 getattr(cube[1], cube_data_prop), threads=threads, verbose=verbose
             )
+            
+            if len(results) == 3:
+                img_cutout, uncert_cutout, aperture = results[0], results[1], results[2]
+            elif len(results) == 2:
+                img_cutout, aperture = results[0], results[1]
 
             # Get cutout wcs info
             cutout_wcs_full = self._get_full_cutout_wcs(cube[2].header)
@@ -912,7 +920,10 @@ class CutoutFactory():
             cutout_wcs_dict = self._get_cutout_wcs_dict()
     
             # Build the TPF
-            tpf_object = self._build_tpf(self.product, cube, img_cutout, uncert_cutout, cutout_wcs_dict, aperture)
+            if self.product == "SPOC":
+                tpf_object = self._build_tpf(cube, img_cutout, cutout_wcs_dict, aperture, uncert_cutout)
+            else:
+                tpf_object = self._build_tpf(cube, img_cutout, cutout_wcs_dict, aperture)
 
             if verbose:
                 write_time = time()
--- a/astrocut/cube_cut.py
+++ b/astrocut/cube_cut.py
@@ -42,6 +42,7 @@ class CutoutFactory():
         Initialization function.
         """
 
+        self.product = None
         self.cube_wcs = None  # WCS information from the image cube
         self.cutout_wcs = None  # WCS information (linear) for the cutout
         self.cutout_wcs_fit = {
@@ -465,7 +466,7 @@ class CutoutFactory():
         return img_cutout, uncert_cutout, aperture
 
 
-    def _update_primary_header(self, product, primary_header):
+    def _update_primary_header(self, primary_header):
         """
         Updates the primary header for the cutout target pixel file by filling in 
         the object ra and dec with the central cutout coordinates and filling in
@@ -473,11 +474,8 @@ class CutoutFactory():
         as we do not have access to this information.
         This is a TESS-specific function.
 
-        Parameters
+        Parameter(s)
         ----------
-        product : str
-            The product type to make the cutouts from.
-            Can either be 'SPOC' or 'TICA'.
         primary_header : `~astropy.io.fits.Header`
             The primary header from the cube file that will be modified in place for the cutout.
         """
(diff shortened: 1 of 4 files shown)


astrocut/cube_cut.py
```diff
@@ -668,15 +672,12 @@ def _apply_header_inherit(self, hdu_list):
...
         Building the cutout target pixel file (TPF) and formatting it to match TESS pipeline TPFs.
 
         Paramters
         ---------
-        product : str
-            The product type to make the cutouts from.
-            Can either be 'SPOC' or 'TICA'.
         cube_fits : `~astropy.io.fits.hdu.hdulist.HDUList`
             The cube hdu list.
         img_cube : `numpy.array`
```
falkben, line 683: (this comment is for line number 685, GH doesn't allow comments on collapsed sections of code)

`uncert_cube` docstring needs to be updated

in general, we should update the docstrings everywhere we have changed the signature of the functions

astrocut/cube_cut.py
```diff
@@ -845,6 +847,9 @@ def cube_cut(
         if verbose:
             start_time = time()
 
+        # Declare the product type being used to make the cutouts
+        self.product = product
```
falkben, line 851: why is `product` an argument of cube_cut() function still? 

self.product should only be assigned during initialization of the cutout factory, now, right?

astrocut/cube_cut.py
```diff
@@ -895,9 +900,14 @@ def cube_cut(
                 print("ymin,ymax: {}".format(self.cutout_lims[0]))
 
             # Make the cutout
-            img_cutout, uncert_cutout, aperture = self._get_cutout(
+            results = self._get_cutout(
                 getattr(cube[1], cube_data_prop), threads=threads, verbose=verbose
             )
+            
+            if len(results) == 3:
+                img_cutout, uncert_cutout, aperture = results[0], results[1], results[2]
```
falkben, line 908:
```suggestion
                img_cutout, uncert_cutout, aperture = results
```

astrocut/cube_cut.py
```diff
@@ -895,9 +900,14 @@ def cube_cut(
...
+            elif len(results) == 2:
+                img_cutout, aperture = results[0], results[1]
```
falkben, line 910:
```suggestion
                img_cutout, aperture = results
```

astrocut/cube_cut.py
```diff
@@ -447,45 +448,48 @@ def _get_cutout(self, transposed_cube, threads: Union[int, Literal["auto"]] = 1,
...
             aperture = np.pad(aperture, padding[1:], 'constant', constant_values=0)
 
         if verbose:
             print("Image cutout cube shape: {}".format(img_cutout.shape))
-            print("Uncertainty cutout cube shape: {}".format(uncert_cutout.shape))
+            if self.product == "SPOC":
+                print("Uncertainty cutout cube shape: {}".format(uncert_cutout.shape))
     
-        return img_cutout, uncert_cutout, aperture
+        if self.product == "SPOC":
```
falkben, line 469: returning different amounts of things instead of always the same amount of things makes logic of dealing with the return values much harder.

the logic of dealing with spoc vs. tica is spreading throughout the codebase.

what do you think about initializing `uncert_cutout` to `None` and return all three variables no matter what?

we also need to update the docstring about the return value with whatever this function ends up doing.

astrocut/cube_cut.py
```diff
@@ -668,15 +672,12 @@ def _apply_header_inherit(self, hdu_list):
                         hdu.header[kwd] = (primary_header[kwd], primary_header.comments[kwd])
             
 
-    def _build_tpf(self, product, cube_fits, img_cube, uncert_cube, cutout_wcs_dict, aperture):
+    def _build_tpf(self, cube_fits, img_cube, cutout_wcs_dict, aperture, uncert_cube=None):
```
falkben, line 675: if when we make the cutout, it returns `None` for `uncert_cube`, we can call this exactly the same for both TICA and SPOC and we can leave uncert_cube as a required param. then we don't need another conditional when we call this function, since for TICA `None` will be passed in.

astrocut/make_cube.py
```diff
@@ -440,7 +440,7 @@ def _configure_cube(self, file_list, **extra_keywords):
         # Determining cube shape: 
         # If it's a new cube, the shape is (nRows, nCols, nImages, 2)
```
falkben, line 441: this comment should be updated, maybe include something about the reason it is 1 instead of 2

falkben (changes requested):
tesscut code will need to change for this to work

astroquery? anything else?

astrocut/cube_cut.py
jaymedina: `self.product` is assigned to `None` when the `CutoutFactory` class is initialized. It's then replaced by "TICA" or "SPOC" depending on what value the user passes when calling `cube_cut`. But this reminds me, I should enforce that the "TICA" or "SPOC" string is in all uppercase so we don't run into a case-sensitive bug down the line.

jaymedina:
> tesscut code will need to change for this to work
> 
> astroquery? anything else?

`astroquery.mast.Tesscut` wraps around the TESSCut API, so once we redeploy TESSCut API, we should be good to go (I will keep an eye on unit tests). We also have to ensure that [this notebook](https://github.com/spacetelescope/mast_notebooks/blob/main/notebooks/astrocut/making_tess_cubes_and_cutouts/making_tess_cubes_and_cutouts.ipynb) runs correctly, too. More detail is in the [Jira ticket](https://jira.stsci.edu/browse/ASB-22309) for this Epic.

astrocut/cube_cut.py
falkben: i see, i missed that originally -- i guess i'd expected it to be an argument to the initialization of the coutoutfactory instance.

in general, i think this changeset is moving pretty far away from the idea of a factory class and we're now adding a bunch of conditionals if / else throughout the codebase which i'm not a huge fan of. 

perhaps we should hold off on refactoring the product and first just implement the change to remove the error array from TICA cubes?

When refactoring, i think there might be some ways to try and avoid all the conditional logic that is in this PR, which gets very hard to maintain.

jaymedina: We've already removed the error array from the TICA cubes in PR #87 

I ran some tests and confirmed that the removal of the error arrays from the TICA cubes is worth the hassle of refactoring things in `CutoutFactory` (cubes are now 1/2 the size and take about 1/3 less time to make). I thought about making a new class just for TICA called `TicaCutoutFactory`, the same way things exist separately in `make_cube.py`, but it leads to a lot of redundancies.

falkben: sorry, i mean cutouts, not cubes, above.

i'm not disagreeing about removing the error component from the cutout (or the cube). 

i wonder, do fits files support sparse arrays?

jaymedina: You can make sparse matrices as numpy arrays, so I assume if `fits` can support a numpy array, they can support a sparse array if numpy is its data type.

falkben: I count 11 conditionals where we're checking if the cutout product is TICA or SPOC when making a cutout and I'm worried about the added complexity that has been added to this codebase by adding just a single new product type.

What I would like is a lot less conditional behavior based on the product and to instead utilize the factory based structure this module purports to be to avoid all these conditionals. 

We can handle this in a separate PR / ticket if you like. I do think this is important for maintainability, both for the cube creation as well as the cutouts.  I did mention some of this when we added TICA to the cube creation module as well.

With some refactoring, we can possibly simplify this to use the factory methodology and define objects (SPOC / TICA) that use the factory to make the cutouts, and can keep the different behavior isolated to their own class methods.  If you like, we can work together to sketch this out.

I think we should not undertake a large refactoring in this PR, so I will instead focus my comments on the specifics of not returning the error array when making TICA cutouts.

astrocut/cube_cut.py
```diff
@@ -403,7 +404,7 @@ def _get_cutout(self, transposed_cube, threads: Union[int, Literal["auto"]] = 1,
         -------
         response :  `numpy.array`, `numpy.array`, `numpy.array`
```
falkben, line 405: should this line change as well? the second item is either a numpy array or None.

some type annotations on these functions would also help a lot. i think it would be valuable to add mypy at some point to this project and to type annotate as much as we can.

astrocut/tests/test_cube_cut.py
```diff
@@ -362,12 +362,11 @@ def test_taget_pixel_file(cube_file, ffi_type, tmp_path):
...
 
     tpf_table = tpf[1].data
-    # SPOC cutouts have one extra column in EXT 1
-    ncols = 12 if ffi_type == 'SPOC' else 11
+    # SPOC cutouts have 2 extra columns in EXT 1
+    ncols = 12 if ffi_type == 'SPOC' else 10
     assert len(tpf_table.columns) == ncols
     assert "TIME" in tpf_table.columns.names
     assert "FLUX" in tpf_table.columns.names
-    assert "FLUX_ERR" in tpf_table.columns.names
```
falkben, line 370: can we check that FLUX_ERR is present for SPOC and not present in TICA?

falkben (changes requested):

astrocut/cube_cut.py
jaymedina: I see what you mean, and yes it makes sense to do this refactoring in a separate PR. This PR has a smaller time-constraint, as we have to start generating the errorless TICA cubes as soon as Aug. 14th, so I'd like to have this merged by the end of this week (Aug. 3rd) ideally. That way @hmlewis-astro can start generating the error-free TICA cubes asynchronously while I work on simplifying the code. Let me know if you agree with this and I'll make a new Jira ticket for the refactoring @falkben

falkben: That sounds good. 

Also, this might be related: https://github.com/spacetelescope/astrocut/issues/62 since there was a similar discussion when the make cube stuff was added for TICA https://github.com/spacetelescope/astrocut/pull/59#discussion_r890526958

I'm not sure what makes sense re: refactoring, whether class inheritance would actually make things more maintainable etc, but we might want to have these work items linked, somehow.  

I'd be happy to sketch this out with you at some point to try and generalize this a bit.

jaymedina: OK, I will link all three PRs (#88, #62, and #59) in the Astrocut Refactoring Jira ticket. Thank you!

jaymedina:
> There's some flake8 failures in CI -- look relatively trivial to fix: https://github.com/spacetelescope/astrocut/actions/runs/5716882528/job/15489424621?pr=88

Interesting - I haven't seen that codestyle error before. I'll update with a new commit.

falkben (dismissed):
regarding the [CI failure](https://github.com/spacetelescope/astrocut/actions/runs/5717424375/job/15491139878?pr=88) for python 3.11 and installing everything from source, I think we might need an additional step to install openblas: `libblas-dev`

this can be done in a different PR, though. Once this is merged, I don't mind taking a look.

falkben (approved):

jaymedina:
Before we merge this into main: 

I demo'ed the work here earlier this week and got a suggestion from @scfleming to keep the `FLUX_ERR` cutouts in TICA for the sake of making the code easier to integrate into 3rd party software like `Lightkurve` and user scripts. I believe this would involve changing [the logic that starts here](https://github.com/spacetelescope/astrocut/blob/7d488c5842144054adae76d1b4099006d8b3a3d9/astrocut/cube_cut.py#L733) in this branch so that an empty array (`empty_arr`) gets filled in for each row of this column when the product is TICA, and also updating the unit tests accordingly (we will check TICA for the existence of `FLUX_ERR` again, and ensure that they are empty arrays). Let me know what you all think of this logic before I make a commit @falkben @hmlewis-astro

falkben:
I'd be happy with that change @jaymedina -- it will make this cutout code easier to maintain, as well.

It will make the cutouts the same size as the SPOC cutouts, though, (not any smaller) but I don't think that was the intention with any of this.

jaymedina:
> It will make the cutouts the same size as the SPOC cutouts, though, (not any smaller) but I don't think that was the intention with any of this.

Updated accordingly. Yep, the cutouts themselves are small, so it's not as big of a deal for them to have the error array. The important thing was to reduce the size of the cubes by removing the error array _there_, and to adapt the `CutoutFactory` code to take this change into account. Thanks for the quick response!

astrocut/cube_cut.py
```diff
@@ -729,10 +729,10 @@ def _build_tpf(self, cube_fits, img_cube, uncert_cube, cutout_wcs_dict, aperture
 
         # Adding flux and flux_err (data we actually have!)
         pixel_unit = 'e-/s' if self.product == 'SPOC' else 'e-'
+        flux_err_array = uncert_cube if self.product == 'SPOC' else empty_arr
```
falkben, line 732: i'm a little uncertain about this. numpy array assignment is done by reference. i'm just worried that because we are using empty_arry here to assign into cadence_arry, or appending to columns, that something might change inside of `empty_arr` and then would be applied to flux_err_arry at this point.

e.g. to illustrate what i'm worried about

```python
>>> import numpy as np
>>> empty_arr = np.zeros((2,3))
>>> other_arr = empty_arr[:,1]
>>> other_arr[0] = 1
>>> other_arr
array([1., 0.])
>>> empty_arr
array([[0., 1., 0.],
       [0., 0., 0.]])
```

maybe on this line, we should assign flux_err_arry to just `np.zeros(img_cube.shape)` to be sure it's empty?

maybe we're not changing `cadence_array` so this isn't an issue?

falkben: another think i'm wondering -- should the error be zero? or should it be np.NaN?

scfleming: I thought about this too.  I lean towards 0. for two reeasons at the moment:
- the TICA cubes had the dummy values sent to 0. I think, so this maintains that part of it
- adding NaN may result in issues for non-NaN-safe calculations, e.g., produce values of NaN for computation that included the flux uncertainty.  Of course 0. could also introduce problems depending on what the user is doing with it, it's not obious to me which is better if someone isn't aware that these are essentially undefined but use the flux uncertainties in calcuations within their code.

falkben: in that case wouldn't it be better to nullify the results than to silently make the calculation with an incorrect value for the error?

scfleming: that was my old philosophy: use NaN's as a forced flag to indicate to an unaware user that something is not defined, and you either need to mask or add nan-safe options to your function calls.  these days I wonder how many functions filter out NaN values automatically or not.  Do you have thoughts on this @hmlewis-astro @jaymedina ?

hmlewis-astro:
I agree, keeping the `FLUX_ERR` in the TICA cutouts is fine, and will make things easier for users that want to run the same/similar code on cutouts made from SPOC and TICA cubes.

astrocut/cube_cut.py
scfleming: +1 on the NaN's.  I just ask we make sure they are filled in such that they are read in correctly with standard astropy fits reading functions, e.g., we don't want to somehow fill them with NaN's and then have them treated as strings when default astropy functions are called on the fits products.  It's probably as easy as making sure the dtype for the FITS table column is correct and you feed them proper np.nan values I guess, although there's a small part of me that worries that having NaN's as not-strings violates a FITS table rule....of course you could then say "we'll just fill that column as strings that say NaN" but then you'll have different dtypes for TICA vs. SPOC in that column...

jaymedina: Now that Ben mentions this, I'm also thinking people might assume the 0s are not fillers, especially since in the metadata we explicitly put the unit of the error array as being in electrons ([see here](https://github.com/spacetelescope/astrocut/blob/49dee429c4226de58c74d889915b454e0a17a9ab/astrocut/cube_cut.py#L731)). What we could do is instead fill the array with some ridiculously large number like 9999999 so that users know straight away that this arrays for TICA are filler arrays? Just a thought. Otherwise, I'm leaning towards making them NaN arrays now.

scfleming: @jaymedina If we go that route (non-zero but valid number) I would vote for a rediculously large NEGATIVE number (-9999999.0 or something), since negative flux uncertainties would also be treated as suspcious, but in theory you could have crazy large positive values for uncertainties.

jaymedina: I'm in favor of that. I'll go ahead and make a commit unless there are other suggestions!

falkben: I am not in favor of using any actual number here -- the error/uncertainty is unknown. using actual numbers is likely just going to result in confusion. if we're returning an array to conform with some expectation that the cutout has a certain number of extensions or whatever, but the values are not to be used -- well, this is precisely what np.NaN is for. we might also consider writing something in the header of the fits file about it?

jaymedina: I'm not opposed to changing the vals to `np.nan`s. One thing to consider tho, is that `np.zero` arrays are used for all the other columns which we don't have information about, for example `CADENCEN0` which is the column that was referenced in the start of this thread (when making references to `cadence_array`). So if 1 column uses `np.nan` to fill in values we don't know about, and 1 column uses `np.zero`, that will be even more confusing.

What we can do for now is stick with the traditional `np.zero` array (I can reassign `flux_err_array` to `np.zeros(img_cube.shape)` to ensure its a zero array as suggested in the first comment of the thread), and once we come to can agreement on what the fill vals should be, I'll open up a new PR to update the `empty_arr` for all relevant columns.

As for timeline, I'd like to get this PR merged in by end of this week or beginning of next week so I have enough time to update and redeploy the TESSCut API before the end of this sprint.

hmlewis-astro: I think this solution works for now! Once you make the changes to reassign `flux_err_array` following Ben's original comment, I think this is ready to merge

jaymedina: Sounds good, thanks @hmlewis-astro !

astrocut/cube_cut.py
```diff
@@ -710,28 +710,30 @@ def _build_tpf(self, product, cube_fits, img_cube, uncert_cube, cutout_wcs_dict,
...
                                 array=(cube_fits[2].columns[start].array + cube_fits[2].columns[stop].array)/2))
 
-        if product == 'SPOC':
+        if self.product == 'SPOC':
             cols.append(fits.Column(name='TIMECORR', format='E', unit='d', disp='E14.7',
                                     array=cube_fits[2].columns['BARYCORR'].array))
 
         # Adding CADENCENO as zeros for SPOC b/c we don't have this info
-        cadence_array = empty_arr[:, 0, 0] if product == 'SPOC' else cube_fits[2].columns['CADENCE'].array
+        cadence_array = np.zeros(img_cube.shape)[:, 0, 0] if self.product == 'SPOC'\
```
jaymedina, line 723: There are other calls to `error_arr` besides flux_err_array, so I decided to just remove the assignment of error_array to cadence_array entirely. @falkben

astrocut/cube_cut.py
```diff
@@ -895,9 +900,8 @@ def cube_cut(
                 print("ymin,ymax: {}".format(self.cutout_lims[0]))
 
             # Make the cutout
-            img_cutout, uncert_cutout, aperture = self._get_cutout(
-                getattr(cube[1], cube_data_prop), threads=threads, verbose=verbose
-            )
+            img_cutout, uncert_cutout, aperture = self._get_cutout(getattr(cube[1], cube_data_prop), threads=threads,
+                                                                   verbose=verbose)
```
falkben, line 904: we should probably align on what formatting whitespace formatting we're using on this project to avoid unnecessary changes back and forth...

I use [darker](https://github.com/akaihola/darker) which just applies black selectively to my set of changes (instead of on the entire file).

falkben (dismissed):

falkben (dismissed):

jaymedina:
Sorry, re-requesting because your review got dismissed @falkben

falkben (approved):

hmlewis-astro (dismissed):

hmlewis-astro (approved):

hmlewis-astro (approved):

PR #89, opened 2023-08-07
Updating `.readthedocs.yml` with Python3.8

jaymedina:
The Astrocut RTD maintainers received an e-mail notification about adding a `.readthedocs.yml` v2 configuration file to our project, as they will no longer be supporting projects with implicit dependencies. More info can be found in this article: https://blog.readthedocs.com/migrate-configuration-v2/ . 

@ceb8 already built us a `.readthedocs.yml` v2 file, so we are good to go. The reason we actually got that e-mail was because we had the following checkbox checked off in the `Advanced Settings` of the `Astrocut` project, which is the feature that will be deprecated in September 2023:

image

All it needed was to be updated to require python 3.8 instead of 3.7, because 3.7 does not jive with `astropy>=5.2`, which caused our documentation to break for the past 4 months, and is the actual reason for this PR:

**TLDR; This PR updates the `.readthedocs.yml` to build documentation on python `3.8` (as opposed to the original `3.7`), because our minimum `astropy` requirement of `>=5.2` doesn't work with `py37`. This PR also updates the minimum python version requested in the `setup.cfg`.**

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,7 +4,7 @@ build:
   image: latest
 
 python:
-  version: 3.7
+  version: 3.8
   install:
     - method: pip
       path: .
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,7 +1,6 @@
 version: 2
 
 build:
-  image: latest
   os: ubuntu-22.04
   tools:
     python: "3.11"
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,7 +2,7 @@ version: 2
 
 build:
   image: latest
-  os: macos-latest
+  os: ubuntu-22.04
   tools:
     python: "3.11"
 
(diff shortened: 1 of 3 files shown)


falkben (commented):
Default python version for readthedocs is 3.11. 3.8 is going end of life next year. Let's just build the docs using 3.11 instead?  It's also faster -- which I think is better for everyone.

Also -- we probably should be testing the package on 3.11, besides just the currently broken dev only dependencies. And we should probably start testing on 3.12 as well. These could be handled in a separate PR.

falkben (changes requested):
in addition to bumping RTD to 3.11, can we also update this to 3.8

https://github.com/spacetelescope/astrocut/blob/main/astrocut/__init__.py#L13

falkben (approved):
assuming docs are building, looks good to me

falkben (approved):

PR #90, opened 2023-08-10
Update cube and cutout unittests

hmlewis-astro:
Updates to existing unit tests and adding new tests to account for changes in `CutoutFactory` (made in #88). Tests verify that cutouts are being made correctly from the new TICA cubes (without error arrays) and verify known differences in TICA vs. SPOC FFIs, cubes, and cutouts.

Note, there is a placeholder unittest (`test_cube_cut.py::test_tica_cutout_error`) to remind us to go back to look at whether we can/should point the test to an existing TICA cube (one that *does* have the error array and will be among the last cubes to me re-built, e.g., Sector 27) that will be re-built later, to verify that cutouts can still be made from cubes with error array.

### DIFF
--- a/astrocut/tests/test_cube_cut.py
+++ b/astrocut/tests/test_cube_cut.py
@@ -431,9 +431,11 @@ def test_tica_cutout_error(tmp_path):
     Test cutouts created from existing TICA cubes (i.e., those with
     error arrays)
 
-    Write test to check cutouts made from TICA cubes that still have
-    the error array. Needed to verify that CutoutFactory will work on
-    the cubes that have not been remade yet.
+    Write test to check that cutouts can be made from *both* the TICA
+    cubes that still have the error array, as well as cubes that
+    *have* been updated and no longer have the error array. Needed to
+    verify that CutoutFactory will work on the cubes that have not 
+    been remade yet.
     """
 
     tmpdir = str(tmp_path)
@@ -691,6 +693,8 @@ def test_s3_tica_cube_cut(tmp_path: Path):
     2
     >>> cut[0][0].header['CCD']  # doctest: +SKIP
     1
+
+    TODO: Remove test `test_tica_cutout_error` when this test starts working
     """
     # Test case: RR Lyrae star in Sector 27 (Camera 2, CCD 1)
     coord = SkyCoord(299.27269, -67.14491, unit="deg", frame="icrs")
--- a/astrocut/tests/test_cube_cut.py
+++ b/astrocut/tests/test_cube_cut.py
@@ -224,18 +224,6 @@ def test_header_keywords_quality(cube_file, ffi_type, tmp_path):
         assert primary_header['CREATOR'] == 'astrocut'
         assert primary_header['BJDREFI'] == 2457000
 
-        # Get units from BinTableHDU
-        cols = hdulist[1].columns.info('name, unit', output=False)
-        cols_dict = dict(zip(*cols.values()))
-
-        ncols = 12 if ffi_type == 'SPOC' else 11
-        assert len(cols_dict) == ncols
-
-        if ffi_type == 'SPOC':
-            assert hdulist[0].header['TIMEREF'] == 'SOLARSYSTEM', 'TIMEREF keyword does not match expected'
-            assert hdulist[0].header['TASSIGN']== 'SPACECRAFT', 'TASSIGN keyword does not match expected'
-            assert cols_dict['FLUX'] == 'e-/s', f'Expected `FLUX` units of "e-/s", got units of "{cols_dict["FLUX"]}"'
-
         if ffi_type == 'TICA':
             # Verifying DATE-OBS calculation in TICA
             date_obs = primary_header['DATE-OBS']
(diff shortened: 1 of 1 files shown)


jaymedina:
Thanks for this PR - I thought about it a bit more about testing the new `CutoutFactory` changes against an older TICA cube (still with the error array) and I think you can just use a SPOC cube with the TICA header keywords superimposed. Before the removal of the error arrays, the TICA cubes were structurally identical to the SPOC ones, just with different header keywords. If we go this route, your code will look like this: 

```
def test_tica_cutout_error():
     tica_cube_error = CubeFactory() # this is actually a SPOC cube but we're gonna update the headers with TICA 
     tica_cube_no_error = TicaCubeFactory()
     
     with fits.open(tica_cube_error, mode='u') as hdulist:
           tica_cube_error[0].header = fits.getheader(tica_cube_no_error, 0)
           tica_cube_error[1].header = fits.getheader(tica_cube_no_error, 1)
           
           etc...
     # now do the actual testing 
     out_file = CutoutFactory(tica_cube_error)
```

astrocut/tests/test_cube_cut.py
```diff
@@ -390,6 +426,49 @@ def test_target_pixel_file(cube_file, ffi_type, tmp_path):
     tpf.close()
 
 
+def test_tica_cutout_error(tmp_path):
```
falkben, line 429: If possible, in cases like this, i find it nice to write out the correct test now, and mark it with xfail. Then, when the behavior is working, we will notice (the test will start failing, and we can just remove the decorator and have it start to pass).

https://docs.pytest.org/en/7.1.x/how-to/skipping.html

hmlewis-astro: I've filled in the placeholder test now, Let me know what you think!

falkben: okay, just to confirm, instead of using xfail here on a real tica cube, we're mocking the tica cube?

hmlewis-astro: Exactly. Creating a cube with TICA headers, but taking the ImageHDU from a "SPOC" cube that includes the error array.

falkben: Okay, I was suggesting something different with xfail. The reason I was suggesting to use xfail is because it's obvious when the test starts working because the test will fail when the assertions all pass.

I'm fine with whatever approach is done here but there are tradeoffs to each. 

For instance, when mocking these TICA files with SPOC -- it could lead to obsolete tests -- we may never come back to update the tests, and with it not actually testing the real TICA data, we wouldn't know if something were broken.

hmlewis-astro: Ah okay, that makes sense. Do you have any thoughts about the best way to access the current TICA cubes? Ideally (I think) without using astroquery.mast.Tesscut?

falkben: We don't necessarily need to access the TICA files here -- just write something that _will_ be able to access them when they are available on S3. If we know what their paths are going to be.

I'm also fine with what you have here, mocking the TICA file, as long as we do come back and point it at an actual TICA file later.

hmlewis-astro: Okay, I added a new test (test_s3_tica_cube_cut), marked with xfail, that tries to make a cutout from a TICA cube on the S3 bucket. That test will continue to xfail until we have *all* of the TICA cubes up on the cloud; it uses a Sector 27 cube, and that will be the last Sector that gets copied to the S3 bucket.

This test we're talking about here (test_tica_cutout_error) is intended to make sure that we can make cutouts from *both* the TICA cubes that do have error arrays (current) and the TICA cubes that do not have error arrays (coming soon), while we are in the process of updating the TICA cubes, since we expect this to take ~weeks.

falkben: very nice! i'll take another look through this later today.

falkben (dismissed):
I will leave it to others to decide if this tested all the various header differences, but these tests look good to me.

astrocut/tests/test_cube_cut.py
```diff
@@ -243,6 +243,42 @@ def test_header_keywords_quality(cube_file, ffi_type, tmp_path):
...
+            assert hdulist[0].header['TIMEREF'] == 'SOLARSYSTEM', 'TIMEREF keyword does not match expected'
+            assert hdulist[0].header['TASSIGN'] == 'SPACECRAFT', 'TASSIGN keyword does not match expected'
+            assert cols_dict['FLUX'] == 'e-/s', f'Expected `FLUX` units of "e-/s", got units of "{cols_dict["FLUX"]}"'
+            assert np.mean(hdulist[1].data.field('CADENCENO')) == 0.0
+
+        if ffi_type == 'TICA':
+            assert hdulist[0].header['TIMEREF'] is None, 'TIMEREF keyword does not match expected'
+            assert hdulist[0].header['TASSIGN'] is None, 'TASSIGN keyword does not match expected'
+            assert cols_dict['FLUX'] == 'e-', f'Expected `FLUX` units of "e-", got units of "{cols_dict["FLUX"]}"'
+            assert np.mean(hdulist[1].data.field('CADENCENO')) != 0.0
```
jaymedina, line 279: Are we certain that this `CADENCEN0` will always be a non-zero value for TICA?

jaymedina (approved):
These tests look good to me! Just had one question on an assertion re: the `CADENCEN0` header keyword for TICA. Other than that, I think this is ready for merge.

astrocut/tests/test_cube_cut.py
hmlewis-astro: Yes, the `CADENCE` keyword in the TICA FFI headers (which matches the `FFIINDEX` in the SPOC headers) is the number of FFIs since the start of data collection. This number increments by 1 with each FFI observed, so it will also be positive and non-zero.

I'm going to make a small adjustment to make it clear that `CADENCENO` for TICA should always be non-zero with
`assert hdulist[1].data.field('CADENCENO').all() != 0.0`

jaymedina: Thank you!

jaymedina (approved):

PR #92, opened 2023-08-17
explicitly close fsspec openfile (astropy doesn't do it)

falkben:


falkben:
Doesn't seem to be necessary and closing the extra fsspec open file handle doesn't resolve the memory leak we were seeing in tesscut

PR #97, opened 2023-10-02
Use GitHub Actions for publishing new releases to PyPI

jaymedina:
Created a GitHub workflow to build wheels, test package, and publish package to the official PyPI, all triggered by the creation and push of a new tag on @spacetelescope/astrocut.

### DIFF
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -9,8 +9,14 @@ on:
     types:
       - published
 
+permissions:
+  contents: read
+  id-token: write
+
 jobs:
 
+  # TODO: Make it so that this only executes when publish-to-test-pypi executes
+  # TODO: Make it so that the tag gets deleted after publish-to-test-pypi finishes executing
   make_tag:
     runs-on: ubuntu-latest
 
@@ -18,9 +24,7 @@ jobs:
     - name: Make a release tag
       run: git tag -a v0.10.0dev -m "dev release"
 
-  # Release to Test PyPI after every MERGE into @spacetelescope/astrocut/main
-  # TODO: Make it so that this only executes when publish-to-test-pypi executes
-  # TODO: Make it so that the tag gets deleted after publish-to-test-pypi finishes executing    
+  # Release to Test PyPI after every MERGE into @spacetelescope/astrocut/main    
   publish-to-test-pypi:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@v1
     name: Publish to Test PyPI
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -23,7 +23,7 @@ jobs:
     # I.e. A MERGE has been made into @spacetelescope/astrocut/main
     steps:
       - name: Git checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: '0'
       - name: Make the test.pypi release tag
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -19,6 +19,6 @@ jobs:
     if: github.repository_owner == 'spacetelescope' && github.event_name == 'push'
     with:
       test_extras: test
-      test_command: pytest astrocut
+      test_command: pytest --pyargs astrocut
     secrets:
       pypi_token: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -48,6 +48,7 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           repository-url: https://test.pypi.org/legacy/
+          skip_existing: True
 
   # Upload to real PyPI on GitHub Releases.
   release-pypi:
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -23,4 +23,4 @@ jobs:
     if: github.repository_owner == 'spacetelescope' && github.event_name == 'push'
     with:
       test_extras: test
-      test_command: pytest -p no:warngings astrocut/tests/test*.py
+      test_command: pytest astrocut/tests/test*.py
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -33,7 +33,7 @@ jobs:
   release-test-pypi:
     name: Publish in-dev package to test.pypi.org
     environment: release-test-pypi
-    if: github.repository_owner == 'spacetelescope' && github.event_name == 'pull_request'
+    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     needs: build-package
 
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -32,10 +32,11 @@ jobs:
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
-          TAG=$(date +'%Y%m%d%H%M%S')
+          TAG=$(date + '%Y%m%d%H%M%S' + '.dev')
           git tag $TAG
-          echo "new tag made! $TAG"
+          echo "New tag made! $TAG"
           git push origin $TAG
+          echo "Github ref:"
           echo ${{ github.ref }}
 
   # Release to Test PyPI
@@ -48,7 +49,7 @@ jobs:
       # We DON'T want to upload to the official PyPI for this job
       repository_url: https://test.pypi.org/legacy/
       # Let's use the tag we generated wtih make_and_push_tag
-      upload_to_pypi: ${{ startsWith(github.ref, 'refs/tags/') }}
+      upload_to_pypi: 'refs/tags/$TAG'
 
   # Remove the tag and push that deletion to main
   delete_tag:
--- a/.github/workflows/pypi-package.yml
+++ b/.github/workflows/pypi-package.yml
@@ -14,31 +14,51 @@ permissions:
 
 jobs:
 
-  # TODO: Make it so that this only executes when publish-to-test-pypi executes
-  # TODO: Make it so that the tag gets deleted after publish-to-test-pypi finishes executing
-  make_tag:
+  make_and_push_tag:
     runs-on: ubuntu-latest
 
+    # We want a new tag to be made ONLY if we are ready to release to Test PyPI.
+    # I.e. A MERGE has been made into @spacetelescope/astrocut/main
+    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
-    - name: Make a release tag
-      run: echo "test"
+      - name: Make the test.pypi release tag
+        run: |
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+          TAG=$(date +'%Y%m%d%H%M%S')
+      - name: Push the tag
+        run: git push origin $TAG
 
-  # Release to Test PyPI after every MERGE into @spacetelescope/astrocut/main    
+  # Release to Test PyPI
   publish-to-test-pypi:
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@v1
     name: Publish to Test PyPI
-    needs: make_tag
+    needs: make_and_push_tag
 
-    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push' && github.ref == 'refs/heads/main'
     with:
       # We DON'T want to upload to the official PyPI for this job
       repository_url: https://test.pypi.org/legacy/
+      # Let's use the tag we generated wtih make_and_push_tag
+      upload_to_pypi: $TAG
+
+  # Remove the tag and push that deletion to main
+  delete_tag:
+    runs-on: ubuntu-latest
+    # This should only run after push-to-test-pypi successfully executed
+    needs: publish-to-test-pypi
+
+    steps:
+      - name: Remove the test.pypi release tag
+        run: |
+          git tag -d $TAG
+          git push origin --delete $TAG
 
   # Release to PyPI after a RELEASE gets PUBLISHED in @spacetelescope/astrocut/main
   publish-to-pypi:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -74,5 +74,13 @@
...
+
+def read_file(*parts):
+    with open(os.path.join(current_path, *parts), encoding='utf-8') as reader:
+        return reader.read()
+
 setup(use_scm_version={'write_to': os.path.join('astrocut', 'version.py'),
-                       'write_to_template': VERSION_TEMPLATE})
+                       'write_to_template': VERSION_TEMPLATE},
+     long_description=read_file('README.rst'),
+     long_description_content_type='text/markdown')
```
falkben, line 86: does this work? it's not markdown, but restructured text.

it should be: `text/x-rst`

from: https://packaging.python.org/en/latest/specifications/core-metadata/#description-content-type

jaymedina: Ah, hmm, that didn't give me any problems, but that's probably because it hasn't gotten to the description part yet. I'll update accordingly, thanks

pllim:
Hello. Since you pinged me...

* I don't think you can push dirty version tag to PyPI, test or not.

```
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
'0.10.dev79+g75abbb8' is an invalid value for Version. Error: Can't use PEP 440 local versions.
```

* Might be overkill to push to PyPI's test server on every commit. You are going to overload them and also exceed your storage quota really fast. Usually people run everything but the upload part unless they are releasing for real (so no need for PyPI's test server).
* Why not use https://github.com/OpenAstronomy/github-actions-workflows ? They have a workflow for publishing.
* Alternately you can also use a label to selectively run build (except upload) on PRs. For example, no need to waste the CI on trivial fixes unrelated to the build machinery. Example: https://github.com/astropy/astropy/blob/e16a7b887aa9592186f467604dc9555291ee801c/.github/workflows/publish.yml#L36

jaymedina:
Thanks! I think for now I'm going to opt-out of selectively running the build based on a specific tag, because I worry that it will lead to PRs accidentally getting merged without a proper build if someone forgets to tag it. I'd rather automatically run build and lint for each commit, and it doesn't seem computationally expensive. I've changed execution of the release to `test.pypi` to go from every commit to every merge to `main`. @falkben let me know what your thoughts are on these cadences. If there are differing opinions I'm happy to reconsider. I'm going to mark this PR as ready-for-review now, and we can begin iterating on this.

falkben:
For creating releases on PyPI, I think the workflow we had been talking about was to upload a test release to TEST PyPI on merge to main and uploading an actual release to PyPI only when creating a tag in the repo. 

For reference, I believe this was the example that I suggested we follow: https://github.com/python-attrs/attrs/blob/main/.github/workflows/pypi-package.yml

However, there may be some complications with pushing a release to TEST pypi on main without creating a tag, due to `setuptools_scm`. 

We probably don't want to push a new tag to this repo for every merge to main, but maybe if a tag was created "locally" within the github workflow where it's attempting to push to TEST PyPI, perhaps then it would work. These would be throwaway releases, of course.

Probably the workflow that @pllim suggested can handle this, or something similar, as it's likely been designed around using `setuptools_scm` whereas the other workflow doesn't use that tool for packaging/versioning.

It would probably be good to try to use the astronomy community's tooling (e.g. the open astronomy workflows) rather than the more general workflow that I originally suggested. (I didn't know about OpenAstronomy workflows until just now)

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,22 @@
+name: Building & Publishing
+
+on:
+  push:
+    tags: ["v*"]
```
jaymedina, line 5: I specified that this workflow should only trigger after tags starting with `v` are pushed (i.e. `v0.10.0`)

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  # Release to PyPI after a RELEASE gets PUBLISHED in @spacetelescope/astrocut/main
+  publish-to-pypi:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    name: Publish to PyPI
+
+    # We only want to upload to PyPI after a new release has been published on the @spacetelescope/astrocut repo
+    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push'
+    with:
+      test_extras: test
```
jaymedina, line 21: I verified that the `extras-require` in our `tox.ini` for testing is called `test`

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,22 @@
...
+      test_command: pytest astrocut/tests/test*.py
```
jaymedina, line 22: This is the command I run in a terminal (from the base directory of my local clone of `astrocut` repo) to get tests to run. I have also seen an example [where warnings are ignored](https://github.com/spacetelescope/acstools/blob/c01ae65bff403d9b5b6649a9f58bd3d1712fdca1/.github/workflows/publish-to-pypi.yml#L39C41-L39C41) but I opted to leave warnings in, in case we want to see if they change etc. Running these tests on my local machine gives me over 200 warnings.

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,22 @@
+name: Building & Publishing
+
+on:
+  push:
+    tags: ["v*"]
+
+permissions:
+  contents: read
```
jaymedina, line 8: I reverted this back to only `read` permissions so that GitHub Actions does not have the ability to make commits, etc.

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,26 @@
+name: Building & Publishing
+
+on:
+  push:
+    tags: 
+    - "v*"
+    - '!*dev*'
+    - '!*pre*'
+    - '!*post*'
```
jaymedina, line 9: I _don't_ want the `publish-to-pypi` job to run for tags that are pushed with `dev` `pre` or `post` in the name. Just official releases.

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,26 @@
+name: Building & Publishing
+
+on:
+  push:
+    tags: 
+    - "v*"
```
jaymedina, line 6: I specified that this workflow should only trigger after tags starting with v are pushed (i.e. v0.10.0)

.github/workflows/pypi-package.yml
falkben: why not? these can be useful for testing and preparing a release. I would suggest that once this PR is merged we create a dev tag to check that this workflow is running.

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,26 @@
...
+  # Release to PyPI after a RELEASE gets PUBLISHED in @spacetelescope/astrocut/main
+  publish-to-pypi:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    name: Publish to PyPI
+
+    # We only want to upload to PyPI after a new release has been published on the @spacetelescope/astrocut repo
+    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push'
+    with:
+      test_extras: test
+      test_command: pytest astrocut/tests/test*.py
```
falkben, line 26: Don't we need:

```yml
    secrets:
      pypi_token: ${{ secrets.PYPI_PASSWORD_STSCI_MAINTAINER }}
```

in here? (or whatever the secret is called)

.github/workflows/pypi-package.yml
falkben: I'm not seeing the `id-token: write` perm on any of the other ST workflows. Can we remove it?

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,26 @@
...
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+
+  # Release to PyPI after a RELEASE gets PUBLISHED in @spacetelescope/astrocut/main
+  publish-to-pypi:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    name: Publish to PyPI
```
falkben, line 20: I'm not sure 'name' is supported here?

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,26 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+
+  # Release to PyPI after a RELEASE gets PUBLISHED in @spacetelescope/astrocut/main
+  publish-to-pypi:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
```
falkben, line 19: we're not defining `upload_to_pypi` anywhere which I believe needs to be set to `true` or a conditional that evaluates to `true` in order for this to actually publish it to pypi

https://github.com/OpenAstronomy/github-actions-workflows/blob/31c0e47a060402a9bd2d97863ea50577d509c050/.github/workflows/publish_pure_python.yml#L126C13-L136

.github/workflows/pypi-package.yml
falkben: perhaps we should make a ticket to address some of these warnings?  if the warnings are not ignored, will the cause the tests to fail?

setup.py
```diff
@@ -74,5 +74,13 @@
...
+
+def read_file(*parts):
+    with open(os.path.join(current_path, *parts), encoding='utf-8') as reader:
+        return reader.read()
+
 setup(use_scm_version={'write_to': os.path.join('astrocut', 'version.py'),
-                       'write_to_template': VERSION_TEMPLATE})
+                       'write_to_template': VERSION_TEMPLATE},
+     long_description=read_file('README.rst'),
+     long_description_content_type='text/x-rst')
```
falkben, line 86: note that this can fail if there are errors parsing the rst file. we can address if they come up later, just mentioning here so we are prepared.

falkben (changes requested):

.github/workflows/pypi-package.yml
jaymedina: It looks like there are 2 tokens - I'm guessing we just need the password, based off two examples here:

1. https://github.com/spacetelescope/gwcs/blob/e82d3fb937215c3b18582652740918a88ed9caa7/.github/workflows/build.yml#L15

2. https://github.com/spacetelescope/romancal/blob/a939846a68ed4362f50d6ae4c9317f7877a362f1/.github/workflows/build.yml#L23

I'll update accordingly.

.github/workflows/pypi-package.yml
jaymedina: I have the GitHub Actions extension installed on VSCode and it underlines things that are not supported/formatted correctly. This `name` arg isn't underlined, and it also worked fine when I was running the Test PyPI jobs.

.github/workflows/pypi-package.yml
jaymedina: I removed that arg because it defaults to `/refs/tags/v` according to their code, which is what I wanted:

https://github.com/OpenAstronomy/github-actions-workflows/blob/31c0e47a060402a9bd2d97863ea50577d509c050/.github/workflows/publish_pure_python.yml#L34

Which is the same as feeding it a `${{ startsWith: github.ref, /refs/tags/v }}` (wrong syntax but you get the point):

https://github-actions-workflows.openastronomy.org/en/stable/publish_pure_python.html

.github/workflows/pypi-package.yml
jaymedina: I was wondering the same thing. When testing is commanded, [this action](https://github.com/OpenAstronomy/build-python-dist/blob/main/action.yml) is used, but I looked through that repo and don't see any clues about whether warnings are considered failures. For now I'm assuming that, since our CI builds currently pass even though we get warnings, that running the same tests in this action will pass too?

.github/workflows/pypi-package.yml
```diff
@@ -0,0 +1,24 @@
...
+    name: Publish to PyPI
+
+    # We only want to upload to PyPI after a new release has been published on the @spacetelescope/astrocut repo
+    if: github.repository_owner == 'spacetelescope' && github.event_name == 'push'
+    with:
+      test_extras: test
+      test_command: pytest astrocut/tests/test*.py
```
falkben, line 22: The docs for this mention that this is run from an empty directory so I guess we'll see whether this works

https://github.com/OpenAstronomy/build-python-dist#build-a-source-distribution-with-testing

falkben (commented):
Have we tested that this works? I think creating a tag from the latest commit hash in this branch might get this action to run?

.github/workflows/pypi-package.yml
falkben: I think you might want the following:

```suggestion
      test_command: pytest --pyargs astrocut
```

falkben (changes requested):
Would be good to update the readme with new instructions for creating a new release

README.rst
```diff
@@ -90,9 +90,19 @@ The built docs will be in ``docs/_build/html/``, to view them go to ``file:///pa
...
-^^^^^^^^^^^^^^^^
+================
 
-Follow the `Astropy template release instructions <https://docs.astropy.org/en/latest/development/astropy-package-template.html>`_. 
+GitHub Action Releases
+^^^^^^^^^^^^^^^^^^
+
+The `pypi-package.yml` GitHub workflow is built to make a package release off the latest commit hash in `main`. The job in this workflow is triggered when a
+release tag is generated and pushed to `main`, and uses `OpenAstronomy`'s `GitHub action workflow <https://github.com/OpenAstronomy/github-actions-workflows>`_
+for publishing pure Python packages (`see here <https://github-actions-workflows.openastronomy.org/en/stable/publish_pure_python.html>`_ for documentation).
```
falkben, line 100:
```suggestion
The `pypi-package.yml <.github/workflows/pypi-package.yml>`_ GitHub workflow creates a PyPI release. The job in this workflow is triggered when a tag is pushed or a GH release (+tag) is created, and uses `OpenAstronomy`'s `GitHub action workflow <https://github.com/OpenAstronomy/github-actions-workflows>`_
for publishing pure Python packages (`see here <https://github-actions-workflows.openastronomy.org/en/stable/publish_pure_python.html>`_ for documentation).
```

falkben (changes requested):

README.rst
falkben: trying to write rst links by hand so i might have gotten the formatting wrong there ^^

falkben (approved):

PR #99, opened 2023-10-06
Add Test PyPI action to GitHub Publishing Workflow

### LINKED ISSUE #102: GitHub Action to relase to Test PyPI for each merge

jaymedina:
Closes #102

- [x] Successful upload to Test PyPI with Twine instead of OpenAstronomy action
- [x] Trigger run after being labeled `publish-to-test`
- [ ] Incrementally increase version number based on what already exists in Test PyPI

### DIFF
--- a/.github/workflows/test-pypi-package.yml
+++ b/.github/workflows/test-pypi-package.yml
@@ -15,9 +15,8 @@ jobs:
         # We want a new tag to be made ONLY if we are ready to release to Test PyPI.
         # I.e. A MERGE has been made into @spacetelescope/astrocut/main
         steps:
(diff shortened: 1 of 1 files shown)


jaymedina:
Let me know what you think of the versioning for this @falkben . Ideally, I would have liked to keep the versioning syntax that's used in the official `pypi` releases, but that seemed tricky to do, although I didn't explore the solutions enough. I went with defining each version based on the date/time of the current run, which is simple. Users CAN choose to download from Test PyPI so I wonder if we should keep our version numbers the same as how we write them in PyPI.

PR #100, opened 2023-10-06
"Decommissioning" `update_cube` for now

jaymedina:
`update_cube` -> `_update_cube`


falkben (changes requested):
This method is used elsewhere in the codebase and needs to be updated there. This is why the CI tests are failing. vscode has refactoring tools that will update the names in the project automatically, but it should be simple enough even with find/replace.

PR #105, opened 2023-10-18
New Feature: ASDF Cutout Functionality

jaymedina:
## Background

The Roman Space Telescope will have its data stored in a new file format called [ASDF](https://asdf.readthedocs.io/en/latest/). In order to continue supporting our userbase and enabling productive science, the `astrocut` tool needs to be expanded to allow for the creation of cutouts of Roman mission products.

## Changes

- [x] Enable ASDF file format cutout functionality in `astrocut` with a proof-of-concept script (`asdf_cutouts.py`)
- [ ] Generalize the `asdf_cutouts.py` enough such that it can "merged" into the `cutouts.py` script
- [ ] Unit tests (DON'T UPLOAD ROMAN DATA TO REPO)
- [ ] Documentation update
- [x] Dependency updates

## Preview

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -14,12 +14,12 @@ def get_center_pixel(file, ra, dec):
 
         # Get the WCS header
         header = wcs.to_fits_sip()
-    
+
     # Update WCS header with some keywords that it's missing.
     # Otherwise, it won't work with astropy.wcs tools (TODO: Figure out why. What are these keywords for?)
     for k in ['cpdis1', 'cpdis2', 'det2im1', 'det2im2', 'sip']:
         header[k] = 'na'
-    
+
     # New WCS object with updated header
     wcs_updated = astropy.wcs.WCS(header)
 
@@ -33,13 +33,12 @@ def get_center_pixel(file, ra, dec):
 
 
 def get_cutout(file, coords, wcs, size=20, outfile="example_roman_cutout.fits"):
-    
+
     # Get the 2D science image
     with asdf.open(file) as f:
         data = f['roman']['data']
 
-        coordinates = coords
-        cutout = astropy.nddata.Cutout2D(data, position=coordinates, wcs=wcs, size=(size, size))
+        cutout = astropy.nddata.Cutout2D(data, position=coords, wcs=wcs, size=(size, size))
 
         astropy.io.fits.writeto(outfile, data=cutout.data.value, header=cutout.wcs.to_header(), overwrite=True)
 
@@ -53,4 +52,3 @@ def asdf_cut(input_file, ra, dec, *, cutout_size=20, output_file="example_roman_
     pixel_coordinates, wcs = get_center_pixel(input_file, ra, dec)
 
     get_cutout(input_file, pixel_coordinates, wcs, size=cutout_size, outfile=output_file)
-    
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -8,13 +8,12 @@ import astropy
 def get_center_pixel(file, ra, dec):
 
     # Get the roman 2D science image
-    f = asdf.open(file)
+    with asdf.open(file) as f:
+        # Get the WCS
+        wcs = f['roman']['meta']['wcs']
 
-    # Get the WCS
-    wcs = f['roman']['meta']['wcs']
-    
-    # Get the WCS header
-    header = wcs.to_fits_sip()
+        # Get the WCS header
+        header = wcs.to_fits_sip()
     
     # Update WCS header with some keywords that it's missing.
     # Otherwise, it won't work with astropy.wcs tools (TODO: Figure out why. What are these keywords for?)
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -4,23 +4,25 @@
 
 import asdf
 import astropy
+import gwcs
 
 
-def get_center_pixel(file, ra, dec):
+def get_center_pixel(wcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
     """ Get the center pixel from a roman 2d science image """
 
-    # Get the roman 2D science image
-    with asdf.open(file) as f:
-        # Get the WCS
-        wcs = f['roman']['meta']['wcs']
+    # # Get the roman 2D science image
+    # with asdf.open(file) as f:
+    #     # Get the WCS
+    #     wcs = f['roman']['meta']['wcs']
 
-        # Get the WCS header
-        header = wcs.to_fits_sip()
+    # Get the WCS header
+    header = wcs.to_fits_sip()
 
     # Update WCS header with some keywords that it's missing.
     # Otherwise, it won't work with astropy.wcs tools (TODO: Figure out why. What are these keywords for?)
     for k in ['cpdis1', 'cpdis2', 'det2im1', 'det2im2', 'sip']:
-        header[k] = 'na'
+        if k not in header:
+            header[k] = 'na'
 
     # New WCS object with updated header
     wcs_updated = astropy.wcs.WCS(header)
@@ -34,24 +36,29 @@ def get_center_pixel(file, ra, dec):
     return (row, col), wcs_updated
 
 
-def get_cutout(file, coords, wcs, size=20, outfile="example_roman_cutout.fits"):
+def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: tuple,
+               wcs: astropy.wcs.wcs.WCS, size: int = 20, outfile: str = "example_roman_cutout.fits"):
     """ Get a roman image cutout """
 
-    # Get the 2D science image
-    with asdf.open(file) as f:
-        data = f['roman']['data']
+    # # Get the 2D science image
+    # with asdf.open(file) as f:
+    #     data = f['roman']['data']
 
-        cutout = astropy.nddata.Cutout2D(data, position=coords, wcs=wcs, size=(size, size))
+    cutout = astropy.nddata.Cutout2D(data, position=coords, wcs=wcs, size=(size, size))
 
-        astropy.io.fits.writeto(outfile, data=cutout.data, header=cutout.wcs.to_header(), overwrite=True)
+    astropy.io.fits.writeto(outfile, data=cutout.data, header=cutout.wcs.to_header(), overwrite=True)
 
 
-def asdf_cut(input_file, ra, dec, *, cutout_size=20, output_file="example_roman_cutout.fits"):
+def asdf_cut(input_file: str, ra: float, dec: float, cutout_size: int = 20, output_file: str = "example_roman_cutout.fits"):
     """ Preliminary proof-of-concept functionality.
     Takes a single ASDF input file (``input_file``) and generates a cutout of designated size ``cutout_size``
     around the given coordinates (``coordinates``).
     """
 
-    pixel_coordinates, wcs = get_center_pixel(input_file, ra, dec)
+    with asdf.open(input_file) as f:
+        data = f['roman']['data']
+        gwcs = f['roman']['meta']['wcs']
 
-    get_cutout(input_file, pixel_coordinates, wcs, size=cutout_size, outfile=output_file)
+        pixel_coordinates, wcs = get_center_pixel(gwcs, ra, dec)
+
+        get_cutout(data, pixel_coordinates, wcs, size=cutout_size, outfile=output_file)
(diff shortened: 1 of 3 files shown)


hmlewis-astro (commented):
Thanks, Jenny! This looks like it's in good shape for us to pick-up!

havok2063 (approved):
This looks good to me as well.

havok2063:
Shall we mark this PR as ready to try and get this merged in?

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+def get_center_pixel(file, ra, dec):
+
+    # Get the roman 2D science image
+    f = asdf.open(file)
+
+    # Get the WCS
+    wcs = f['roman']['meta']['wcs']
+    
+    # Get the WCS header
+    header = wcs.to_fits_sip()
```
braingram, line 17:
```suggestion
    # Get the roman 2D science image
    with asdf.open(file) as f:
        # Get the WCS
        wcs = f['roman']['meta']['wcs']

        # Get the WCS header
        header = wcs.to_fits_sip()
```
Using the `with` context here will help to close the opened file.

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+def get_cutout(file, coords, wcs, size=20, outfile="example_roman_cutout.fits"):
+    
+    # Get the 2D science image
+    f = asdf.open(file)
+    data = f['roman']['data']
+
+    coordinates = coords
+    cutout = astropy.nddata.Cutout2D(data, position=coordinates, wcs=wcs, size=(size, size))
+
+    astropy.io.fits.writeto(outfile, data=cutout.data.value, header=cutout.wcs.to_header(), overwrite=True)
```
braingram, line 45:
```suggestion
    # Get the 2D science image
    with asdf.open(file) as f:
        data = f['roman']['data']

        coordinates = coords
        cutout = astropy.nddata.Cutout2D(data, position=coordinates, wcs=wcs, size=(size, size))

        astropy.io.fits.writeto(outfile, data=cutout.data.value, header=cutout.wcs.to_header(), overwrite=True)
```
Similar to the other comment this will help to close the file.

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+    return (row, col), wcs_updated
+
+
+def get_cutout(file, coords, wcs, size=20, outfile="example_roman_cutout.fits"):
+    
+    # Get the 2D science image
+    f = asdf.open(file)
+    data = f['roman']['data']
+
+    coordinates = coords
```
braingram, line 42: Is this assignment needed or can the argument be renamed to `coordinates`?

setup.cfg
```diff
@@ -16,9 +16,12 @@ packages = find:
 python_requires = >=3.8
 setup_requires = setuptools_scm
 install_requires =
+    asdf>=2.15.0 # for ASDF file format
     astropy>=5.2 # astropy with s3fs support
     fsspec[http]>=2022.8.2  # for remote cutouts
     s3fs>=2022.8.2  # for remote cutouts
+    rad>=0.17.0
+    roman_datamodels>=0.17.0
```
braingram, line 24:
```suggestion
    roman_datamodels>=0.17.0 # for roman file support
```

setup.cfg
```diff
@@ -16,9 +16,12 @@ packages = find:
 python_requires = >=3.8
 setup_requires = setuptools_scm
 install_requires =
+    asdf>=2.15.0 # for ASDF file format
     astropy>=5.2 # astropy with s3fs support
     fsspec[http]>=2022.8.2  # for remote cutouts
     s3fs>=2022.8.2  # for remote cutouts
+    rad>=0.17.0
```
braingram, line 23: As `rad` isn't directly used here. What about removing this and allowing `roman_datamodels` to install a suitable `rad` version?

braingram (commented):
Thanks for making these changes. I left a few comments mainly related to using `with` to help closing opened files. Please feel free to reach out with any asdf questions.

Are there plans to add asdf file support to the test suite? If so feel free to ping me on the PR and I can start looking at adding astrocut tests to the asdf downstream CI (to make sure that changes in asdf do not negatively impact astrocut).

astrocut/asdf_cutouts.py
havok2063: I think this makes sense.

setup.cfg
havok2063: Also makes sense to me

havok2063:
@braingram I resolved your comments.   I think we have eventual plans to add asdf support to the test suite.  I'm not sure if those tests should be added now for this code, or if we should wait, since we are planning to fold these changes into the core astrocut code through some refactor and generalization.    But if people want them in now, I can try to add some.

havok2063:
@braingram so the minimum python that astrocut supports is 3.8.  It looks like `roman-datamodels > 0.17` only supports 3.9, while `rad > 0.17` still supports 3.8.  `roman-datamodels > 14` does support 3.8.   I'm not familiar enough with either product.  Given this, would you recommend adding back in the `rad` package dependency or lowering the `roman-datamodels` min version to >14?

falkben:
> @braingram so the minimum python that astrocut supports is 3.8. It looks like `roman-datamodels > 0.17` only supports 3.9, while `rad > 0.17` still supports 3.8. `roman-datamodels > 14` does support 3.8. I'm not familiar enough with either product. Given this, would you recommend adding back in the `rad` package dependency or lowering the `roman-datamodels` min version to >14?

I think we should drop python 3.8. Newer releases of astropy also dropped support of 3.8.

falkben:
> @braingram I resolved your comments. I think we have eventual plans to add asdf support to the test suite. I'm not sure if those tests should be added now for this code, or if we should wait, since we are planning to fold these changes into the core astrocut code through some refactor and generalization. But if people want them in now, I can try to add some.

I think it would be good to add tests here, where we add the functionality.

havok2063:
> I think we should drop python 3.8. Newer releases of astropy also dropped support of 3.8.

Should that be a separate PR?  Is there more to it than bumping the version and removing 3.8 for the CI workflow?  Do we need to coordinate that bump with other related software that use `astrocut` as a dependency?

braingram:
I'll try to provide some context here but I'm not sure I can recommend a solution.

Both `rad` and `roman_datamodels` haven't reached a 'stable' state yet. It's not uncommon to see a breaking change be added that makes opening older files impossible (and similarly opening new files in older versions impossible). Without testing I would not bet on 0.14 opening any current file.

Perhaps `rad` and `roman_datamodels` could be included as [optional requirements](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)? This would allow the main install/CI to retain compatibility and the 'roman' options only tested in a specific job using a newer version of python.

falkben:
> Should that be a separate PR?

I'm fine with it either way. Since it'll be required to go in before this can pass CI, it could be in this or a separate PR that goes in first. If you like I could work on it, but no promises I can get it done quickly.

> Is there more to it than bumping the version and removing 3.8 for the CI workflow? 

Bumping the version throughout the project, making sure the pypi classifiers are correct. i think it's relatively trivial. We should start testing on 3.12 soon. that could also be done at the same time.

> Do we need to coordinate that bump with other related software that use `astrocut` as a dependency?

The only thing I can think of are tesscut. lightkurve calls tesscut over HTTP, so it's not impacted. tesscut is on 3.11 now, so it's not a big deal to drop 3.8. Github doesn't recognize any dependents here: https://github.com/spacetelescope/astrocut/network/dependents anything still on 3.8 can also just install an older version of astrocut. we'll probably not push to pypi immediately, either. or are you thinking of cutting a release soon?

scfleming:
Regarding adding tests, I just want to remind everyone we do not want to release any test Roman data that can be publicly accessed, since there is no official Roman Level 2 test datasets released at MAST yet. Ideally any test would use a purely synthetic asdf file, or can retrieve the file it needs to operate on from a protected source. Or the test itself doesn't go public yet.

havok2063:
> Perhaps `rad` and `roman_datamodels` could be included as [optional requirements](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)? This would allow the main install/CI to retain compatibility and the 'roman' options only tested in a specific job using a newer version of python.

I like the idea of making `roman-datamodels` an optional dependency for now.  Once we migrate this code into the core astrocut code and/or roman data becomes available, we can then make it a required dep.   Since we're dropping 3.8 support, we can consolidate it down to just `roman-datamodel > 0.17`.  

> we'll probably not push to pypi immediately, either. or are you thinking of cutting a release soon?

Saw the 3.8 PR and approved.  I don't think we are planning on releasing soon. 

 
> Regarding adding tests, I just want to remind everyone we do not want to release any test Roman data that can be publicly accessed, since there is no official Roman Level 2 test datasets released at MAST yet. Ideally any test would use a purely synthetic asdf file, or can retrieve the file it needs to operate on from a protected source. Or the test itself doesn't go public yet.

Yeah I would opt for creating a completely fake dataset for the test suite.  We can easily update it as the datamodel changes.  Then once roman data are available and maybe even public, then we add tests with real data.

braingram:
For test data I opened an issue with roman_datamodels to expose something to the public api. There are several 'maker_utils' for generating test data. Here's an example use in jdaviz:
https://github.com/spacetelescope/jdaviz/blob/e8f125ebe375f938ed3a7b2af97e1a1c3dd5b531/jdaviz/configs/imviz/tests/utils.py#L241
It seems more "future-proof" to rely on something added to roman_datamodels as I'm not sure after looking at the jdaviz example if the WCS is kept up-to-date with changes in roman_datamodels.

havok2063:
@falkben @braingram For now I've added some tests with fake data to make sure the basic functionality is there.  Let me know if this seems reasonable to y'all.

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+    with asdf.open(file) as f:
+        # Get the WCS
+        wcs = f['roman']['meta']['wcs']
+
+        # Get the WCS header
+        header = wcs.to_fits_sip()
+
+    # Update WCS header with some keywords that it's missing.
+    # Otherwise, it won't work with astropy.wcs tools (TODO: Figure out why. What are these keywords for?)
+    for k in ['cpdis1', 'cpdis2', 'det2im1', 'det2im2', 'sip']:
```
falkben, line 22: are these always going to be missing? should we check first?

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def get_cutout(file, coords, wcs, size=20, outfile="example_roman_cutout.fits"):
+    """ Get a roman image cutout """
+
+    # Get the 2D science image
+    with asdf.open(file) as f:
+        data = f['roman']['data']
+
+        cutout = astropy.nddata.Cutout2D(data, position=coords, wcs=wcs, size=(size, size))
```
falkben, line 44: should we close the file before starting the cutout? or better yet, can we just pass a file handle in instead of the file path?

astrocut/asdf_cutouts.py
```diff
@@ -0,0 +1,57 @@
...
+        astropy.io.fits.writeto(outfile, data=cutout.data, header=cutout.wcs.to_header(), overwrite=True)
+
+
+def asdf_cut(input_file, ra, dec, *, cutout_size=20, output_file="example_roman_cutout.fits"):
+    """ Preliminary proof-of-concept functionality.
+    Takes a single ASDF input file (``input_file``) and generates a cutout of designated size ``cutout_size``
+    around the given coordinates (``coordinates``).
+    """
+
+    pixel_coordinates, wcs = get_center_pixel(input_file, ra, dec)
```
falkben, line 55: i wonder if it would be better to open the file here, and keep it open, in order to avoid re-opening for getting the center pixel, and then re-opening _again_ to get the data?

falkben (commented):
thanks very much for adding the tests, they look good. just a few minor questions about how we are handling the file/file handle.

astrocut/asdf_cutouts.py
braingram: Good question. In this case the file will need to be kept open until the array data is accessed. By default, asdf lazy-loads array data (only reading the array contents from disk when the array values are accessed). Since the value access doesn't occur until the `Cutout2D` is created I expect that closing the file prior to that point will result in an error when asdf attempts to access the values from the closed file.

astrocut/asdf_cutouts.py
havok2063: I think these are related to handling image distortions.  They might not be common but it's not guaranteed these are always missing.  Adding a check makes sense so we don't accidentally overwrite.

astrocut/asdf_cutouts.py
havok2063: we can consolidate the file IO for sure.

havok2063:
@falkben I've consolidated the file IO into the main function, and added a small header keyword check.  Included some extra cleanup, type annotations and docstrings, etc.

falkben (approved):

PR #109, opened 2023-10-25
Update CHANGES.rst to reflect `v0.10.0` release

jaymedina:
Update CHANGES.rst to reflect `v0.10.0` release

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,35 @@
-0.10 (unreleased)
-----------------
+0.10.1 (unreleased)
+-----------------
+
+
+0.10 (2023-10-23)
+-----------------
 
 - Improve file checking prior to cutout creation to avoid errors [#52]
-- Add remote cutout functionality (requires astropy 5.2 or above) [#76]
+- Fixing broken tests from GitHub Actions CI run [#56]
+- Fixing error resulting from forward slash in target name [#55]
+- MNT: Update codecov-action version to v2 [#53]
+- Make cubes out of TICA FFIs [#59]
+- Make cutouts out of TICA cubes [#60]
+- Fixed bug not catching duplicate ffis [#69]
+- max_memory arg added to update_cube [#71]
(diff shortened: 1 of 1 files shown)


falkben:
woops. we should prob. get in the habit of adding to the changelog in our PR's

CHANGES.rst
```diff
@@ -1,8 +1,35 @@
...
+- Adapting CutoutFactory to account for error-less TICA Cubes [#88]
+- Updating .readthedocs.yml with Python3.11 [#89]
+- Update cube and cutout unittests [#90]
+- Update docs to reflect changes in TICA cube format [#93]
+- Cloud functionality for astrocut.fits_cut() [#95]
+- Decommissioning update_cube for now [#100]
+- Use GitHub Actions for publishing new releases to PyPI [#97]
+- Updating deprecated license_file kwd [#103]
+- Updating syntax in README.rst [#104]
+- Updated README [#108]
```
falkben, line 32: I think for a few of these, with these two lines especially so, we could probably add a bit more info in the changelog, or just remove them if we don't think it's important enough to include in the changelog.  Minor typos fixed in the readme probably aren't worth including in the changelog?

jaymedina:
> woops. we should prob. get in the habit of adding to the changelog in our PR's

No need for habits! I'm adding a workflow to ensure that fails when changelog entries aren't added: https://github.com/spacetelescope/astrocut/pull/110

The current template is from `astroquery` and I will modify accordingly when I have some time.

snbianco:
@dr-rodriguez 

I realized that we should probably update `CHANGELOG.rst` for the new release, but it looks like the changes for 0.10.0 were never merged in. Should I continue on this branch/PR? Or is actually making the release the top priority?

dr-rodriguez:
Interesting. We can probably merge this, maybe flagging 0.11 as unreleased and then updating that when we're good to go. I think the description in this document is useful so I wouldn't want to loose what Jenny had written before.

dr-rodriguez (approved):

PR #110, opened 2023-10-25
CI job to check for changelog update

jaymedina:
A GitHub Actions workflow to check if a new changelog entry has been made for the requested changes. Building this workflow off of https://github.com/astropy/astroquery/blob/main/.github/workflows/changelog.yml

Ensure this job only fails when a changelog entry isn't added for a new feature (unit tests/documentation updates don't need changelog entries).


snbianco:
Closing as a duplicate of #164

PR #112, opened 2023-12-21
drop python 3.8 support

falkben:
asdf dependencies don't allow 3.8 (see PR: https://github.com/spacetelescope/astrocut/pull/105)

we can drop 3.8 (and probably should, since astropy has also dropped 3.8 for newer releases)


havok2063 (approved):
Looks good to me!

PR #114, opened 2024-01-17
Bug fix for asdf cutouts with data as astropy quantity

havok2063:
This PR fixes a bug in the asdf cutout code when the input data to the cutout was an Astropy Quantity instead of a straight numpy array.  Now it accepts both and converts to a data array when writing out to the fits file. 

It also now allows for the return of the cutout object for inspection.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -53,7 +53,8 @@ def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
 
 
 def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: Union[tuple, SkyCoord],
-               wcs: astropy.wcs.wcs.WCS = None, size: int = 20, outfile: str = "example_roman_cutout.fits"):
+               wcs: astropy.wcs.wcs.WCS = None, size: int = 20, outfile: str = "example_roman_cutout.fits",
+               write_file: bool = True) -> astropy.nddata.Cutout2D:
     """ Get a Roman image cutout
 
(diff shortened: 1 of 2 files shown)


astrocut/asdf_cutouts.py
```diff
@@ -53,7 +53,8 @@ def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
 
 
 def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: Union[tuple, SkyCoord],
-               wcs: astropy.wcs.wcs.WCS = None, size: int = 20, outfile: str = "example_roman_cutout.fits"):
+               wcs: astropy.wcs.wcs.WCS = None, size: int = 20, outfile: str = "example_roman_cutout.fits",
+               write_file: bool = True) -> astropy.nddata.Cutout2D:
```
falkben, line 57: this all looks fine, but it's a little awkward that outfile is required, but is ignored if write_file is False. This means that if someone doesn't want output, they still ahve to supply the `outfile`.

I wonder what you think about making outfile default to None, and remove `write_file` boolean. If `outfile` is supplied, then it will write it to the file, if it's None, it will skip it.

falkben: I see that outfile is not required, in fact. Nevertheless, I think we could get rid of the extra parameter here by defaulting `outfile` to `None` and checking it before writing the file

havok2063: Just seeing these comments.  I kind of think the default should be to write a file since I think that's what most users will want.  I added the `write_file` option so I could just get the cutout for inspection.  I'm not opposed to changing it.  I think that's something we can think about more as we integrate this into the other cutout code in a future PR.

PR #115, opened 2024-01-23
Updates to asdf cutout for edge cases

havok2063:
This PR switches the coordinate to pixel conversion to `gwcs`.  It switches the cutout to mode to `partial` which preserves the original cutout shape when requesting a cutout with partial area outside the image bounds.   It wraps the cutout command in try-except when the requested cutout is completely outside the original image.   It adds tests for making cutouts around poles.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -11,7 +11,7 @@ import numpy as np
 from astropy.coordinates import SkyCoord
 
 
-def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
+def get_center_pixel(gwcsobj: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
     """ Get the center pixel from a roman 2d science image
 
     For an input RA, Dec sky coordinate, get the closest pixel location
@@ -19,7 +19,7 @@ def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
 
     Parameters
     ----------
-    gwcs : gwcs.wcs.WCS
+    gwcsobj : gwcs.wcs.WCS
         the Roman GWCS object
     ra : float
         the input Right Ascension
@@ -33,7 +33,7 @@ def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
     """
 
     # Convert the gwcs object to an astropy FITS WCS header
-    header = gwcs.to_fits_sip()
+    header = gwcsobj.to_fits_sip()
 
     # Update WCS header with some keywords that it's missing.
     # Otherwise, it won't work with astropy.wcs tools (TODO: Figure out why. What are these keywords for?)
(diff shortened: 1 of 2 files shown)


scfleming:
This is only being applied to the asdf function. Is the reason that this is already being done in the non-asdf cutout code, or (as a separate PR) are there improvememts to be made for the code that handles FITS like TESSCut or HAPCut with similar justifications?

astrocut/asdf_cutouts.py
```diff
@@ -6,6 +6,7 @@
 import asdf
 import astropy
 import gwcs
```
falkben, line 8: I noticed in `get_center_pixel` we are redefining `gwcs`. And also on line `154`. Can you fix?

astrocut/asdf_cutouts.py
```diff
@@ -47,14 +48,14 @@ def get_center_pixel(gwcs: gwcs.wcs.WCS, ra: float, dec: float) -> tuple:
...
-    row, col = astropy.wcs.utils.skycoord_to_pixel(coords=coordinates, wcs=wcs_updated)
+    row, col = gwcs.invert(coordinates)
 
     return (row, col), wcs_updated
 
 
 def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: Union[tuple, SkyCoord],
                wcs: astropy.wcs.wcs.WCS = None, size: int = 20, outfile: str = "example_roman_cutout.fits",
-               write_file: bool = True) -> astropy.nddata.Cutout2D:
+               write_file: bool = True, fill_value: int = np.nan) -> astropy.nddata.Cutout2D:
```
falkben, line 58: if default is np.nan (float) then this type annotation is not correct. could be `int | float` though?

Also for `asdf_cut()`

falkben (changes requested):

astrocut/asdf_cutouts.py
havok2063: Good catch!  Done.

astrocut/asdf_cutouts.py
havok2063: Switched to `Union[int, float]` since we're the minimum supported python is 3.9.

havok2063:
@scfleming I was focusing on the Roman asdf work for now.  We have a separate ticket to generalize this code and integrate into the core astrocut code.  My thinking was at that time we would make sure the functionality is consistent for all the cutout codes.

scfleming:
Sounds good, I was wanting to know if these changes mean there could be deficiencies to improve in the current cutouts or was unique to Roman. Siounds like there could be gains to be had generally. I agree with the approach to limit to Roman for now.

astrocut/asdf_cutouts.py
```diff
@@ -75,6 +76,8 @@ def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: Union[tuple, Sk
         the name of the output cutout file, by default "example_roman_cutout.fits"
     write_file : bool, by default True
         Flag to write the cutout to a file or not
+    fill_value: int, by default np.nan
```
falkben, line 79: Should this be int or float? (this is why I don't like putting typing info in docstrings...)

astrocut/tests/test_asdf_cut.py
```diff
@@ -19,38 +20,62 @@ def make_wcs(xsize, ysize, ra=30., dec=45.):
...
     # create transformations
+    # - shift coords so array center is at 0, 0 ; reference pixel
+    # - scale pixels to correct angular scale
+    # - project coords onto sky with TAN projection
+    # - transform center pixel to the input celestial coordinate
     pixelshift = models.Shift(-xsize) & models.Shift(-ysize)
     pixelscale = models.Scale(0.1 / 3600.) & models.Scale(0.1 / 3600.)  # 0.1 arcsec/pixel
     tangent_projection = models.Pix2Sky_TAN()
-    celestial_rotation = models.RotateNative2Celestial(ra, dec, 180.)
+    celestial_rotation = models.RotateNative2Celestial(ra, dec, 180.)  #
```
falkben, line 30: empty comment?

```suggestion
    celestial_rotation = models.RotateNative2Celestial(ra, dec, 180.)
```

hmlewis-astro:
Reviewing these updates now! Is this meant to cover pole cases (ASB-24821) AND image edge cases (ASB-24823)?

havok2063:
@hmlewis-astro It wasn't meant to originally, just pole cases.  But I ended up hitting that use case during the pole work, so I added some tests.  I definitely wasn't being exhaustive in my tests for it though.  I'd say still take a look at the code and tests, and if you think there's more to add to sufficiently address the edge cases ticket, please do.  Or if you think this looks good, we can just make a comment on the ticket.

falkben (approved):

PR #116, opened 2024-02-02
Adds option for cutout output to asdf file

havok2063:
This PR adds the option to output the cutout as an asdf file, instead of a fits file.   The asdf cutout file is a minimal tree structure, mirroring the format of the original roman image, i.e.
```
asdf tree:
 - "roman":
   - "meta":
     - "wcs" - the gwcs of the cutout 
   - "data" - the cutout data
```
The included wcs is a sliced gwcs object from the original input to the cutout shape.  This option was chosen to try to preserve as much as possible the original image transformations.  It's a bit of a hack as there's no built-in way of slicing a gwcs yet, nor does `astropy.Cutout2D` support gwcs input yet.    

It also adds a test for the RA=0 edge case.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -209,7 +209,7 @@ def _write_asdf(cutout: astropy.nddata.Cutout2D, gwcsobj: gwcs.wcs.WCS, outfile:
     sliced_gwcs = _slice_gwcs(gwcsobj, cutout.slices_original)
 
     # create the asdf tree
-    tree = {'roman': {'meta': {'wcs': sliced_gwcs, 'data': cutout.data}}}
+    tree = {'roman': {'meta': {'wcs': sliced_gwcs}, 'data': cutout.data}}
     af = asdf.AsdfFile(tree)
 
     # Write the data to a new file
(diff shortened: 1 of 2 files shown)


astrocut/asdf_cutouts.py
```diff
@@ -112,11 +120,102 @@ def get_cutout(data: asdf.tags.core.ndarray.NDArrayType, coords: Union[tuple, Sk
...
+    gwcsobj : gwcs.wcs.WCS
+        the original gwcs object for the full image
+    outfile : str, optional
+        the name of the output cutout file, by default "example_roman_cutout.asdf"
+    """
+    # slice the origial gwcs to the cutout
+    sliced_gwcs = _slice_gwcs(gwcsobj, cutout.slices_original)
+
+    # create the asdf tree
+    tree = {'roman': {'meta': {'wcs': sliced_gwcs, 'data': cutout.data}}}
```
braingram, line 212:
```suggestion
    tree = {'roman': {'meta': {'wcs': sliced_gwcs}, 'data': cutout.data}}
```
This suggestion will but `data` under `roman` (like in your PR description) instead of `meta`.

havok2063: Thanks.  Good catch!

braingram:
Thanks for the review request! At the moment I can only comment on the asdf usage which looks great to me!

falkben (approved):

PR #117, opened 2024-05-17
Handle an s3 URI in asdf_cut()

snbianco:
If an s3 URI is passed in as the `input_file` parameter to the `asdf_cut` function, open the resource with `s3fs` and use its http URL in `asdf.open()`.

This prevents users from having to open the resource themselves before passing the file into the `asdf_cut` function.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -250,7 +250,7 @@ def asdf_cut(input_file: str, ra: float, dec: float, cutout_size: int = 20,
 
     # if file comes from AWS cloud bucket, get URL
     file = input_file
-    if input_file.startswith('s3://'):
+    if isinstance(input_file, str) and input_file.startswith('s3://'):
         fs = s3fs.S3FileSystem()
         with fs.open(input_file, 'rb') as f:
             file = f.url()
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -9,6 +9,7 @@ import asdf
 import astropy
 import gwcs
 import numpy as np
+import s3fs
 
 from astropy.coordinates import SkyCoord
 from astropy.modeling import models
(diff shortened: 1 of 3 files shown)


dr-rodriguez:
Double check if any of the unit tests need to be updated for this.

astrocut/asdf_cutouts.py
```diff
@@ -247,8 +248,15 @@ def asdf_cut(input_file: str, ra: float, dec: float, cutout_size: int = 20,
         an image cutout object
     """
 
+    # if file comes from AWS cloud bucket, get URL
+    file = input_file
+    if isinstance(input_file, str) and input_file.startswith('s3://'):
```
havok2063, line 253: One thing we may want to do at some point is expand this to support `pathlib.Path` as input, in addition to a string.   I don't think `pathlib.Path` handles s3 urls, but it looks like there are packages that do, e.g. `s3path` or `s3pathlib`.   I don't think we need to do it for this PR, but maybe we can create an issue for the future.

havok2063 (commented):
This looks good to me.  I would agree that we should add some explicit tests for string vs s3 path in `test_asdf_cut.py`.

astrocut/asdf_cutouts.py
snbianco: Cool, will make a new issue for this!

snbianco:
@dr-rodriguez @havok2063 

Trying to write unit tests for this now, and wondering what should be used for the `input_file` parameter since we don't have an s3 link with a working ASDF file yet. Also, should we be putting in a link to this data at all given that this is a public repo? I read on one of the other pull requests that the Roman test data shouldn't be publicly accessible.

Also, I'm not too familiar with `tox`, the testing library that we're using. Does anyone know how to generate a coverage report and/or where to find it?

havok2063:
@snbianco We probably don't need to test a real file.  I think we can get away with a fake file for testing, or if there is a public ST S3 file to use for testing?  If we turn lines 251-257 into a tiny function, we could test that functionality separately from any asdf stuff.  That it returns the correct file or s3 file as input?  That would satisfy me for a test.    

It looks like there are packages for mocking s3 aws services, like `moto`.  I'm not sure if we need that here or not.

astrocut/asdf_cutouts.py
snbianco: Issue is ASB-27085

dr-rodriguez:
I generally recommend mocking external resources rather than relying on pings to public files. At a glance, it looked like your approach was good. Basically the goal is to confirm the functions were called with the parameters you expected; the exact logic of the s3fs functions is ideally something that package has verified already.

havok2063 (approved):
This looks good to me.

dr-rodriguez (approved):

PR #118, opened 2024-05-22
Update docs 0.11

snbianco:
Update documentation to prepare for the 0.11.0 release.

Added to the following sections:

- Astrocut Documentation -> ASDF File Cutouts
- Astrocut File Formats -> ASDF Cutout Files
- Astrocut API


astrocut/asdf_cutouts.py
```diff
@@ -16,40 +16,42 @@
...
+    Get the HTTP URI of a cloud resource from an S3 URI.
 
     Parameters
     ----------
     s3_uri : string
         the S3 URI of the cloud resource
     """
     # create file system
-    fs = s3fs.S3FileSystem(anon=True)
+    fs = s3fs.S3FileSystem()
```
snbianco, line 28: This change was to resolve an issue in RSP. Though not directly related to documentation, I thought I would include it in this PR since it's a small fix.

docs/astrocut/index.rst
```diff
@@ -356,10 +356,56 @@ Note that multithreading is disabled by default.
...
+
+                >>> from astrocut import asdf_cut
+                >>> from astropy.coordinates import SkyCoord
+                >>> from astropy.io import fits
+                
+                >>> input_file = ""  # Path to local ASDF file or URI
+
+                >>> center_coord = SkyCoord("80.15189743 29.74561219", unit='deg')
+                
+                >>> cutout_file = asdf_cut(input_file, center_coord.ra, center_coord.dec, cutout_size=200, output_file="roman-demo.fits")  #doctest: +SKIP
```
dr-rodriguez, line 380: Just curious- what does `#doctest: +SKIP` do in this and other lines?

docs/astrocut/index.rst
```diff
@@ -356,10 +356,56 @@ Note that multithreading is disabled by default.
...
+
+ASDF File Cutouts
+===================
+
+The Nancy Grace Roman Space Telescope will store its data using the Advanced Scientific Data Format (ASDF). With the `asdf_cut` function, users can create cutouts of Roman mission products.
+
+Creating ASDF Cutouts
+----------------------
+
+The function `asdf_cut` performs a cutout of an ASDF file and returns the result as either a FITS file or an ASDF file. The cutout ASDF file format is described `here <file_formats.html#asdf-cutout-files>`__.
```
dr-rodriguez, line 368: Since the output can be either FITS or ASDF, maybe your second example can output an ASDF file to demonstrate if there is anything special for calling it. I assume it's just changing it to be something like roman-demo.asdf

docs/astrocut/index.rst
snbianco: Good question! This tells `doctest` (a module that executes Python code in documentation to verify that it works as intended) to ignore this line. Because I don't specify an actual input file in these calls, they don't execute properly and will throw errors without the `SKIP` directive.

PR #121, opened 2024-07-23
Bugfix for Opening Files from the Cloud

snbianco:
Modified the `_get_cloud_http` method to catch a `PermissionError`. If caught, this indicates that the user is trying to access a private resource, as we do in the Astrocut notebook on the Roman Science Platform. If that is the case, then the `S3FileSystem` is initialized with the `anon` parameter set to `False`.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -32,7 +32,7 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], verbose: bool = False) -> str:
     # check if public or private by sending an HTTP request
     s3_path = S3Path.from_uri(s3_uri) if isinstance(s3_uri, str) else s3_uri
     url = f'https://{s3_path.bucket}.s3.amazonaws.com/{s3_path.key}'
-    resp = requests.head(url)
+    resp = requests.head(url, timeout=10)
     is_anon = False if resp.status_code == 403 else True
     if verbose and not is_anon:
         print(f'Attempting to access private S3 bucket: {s3_path.bucket}')
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -25,8 +25,17 @@ def _get_cloud_http(s3_uri: Union[str, S3Path]) -> str:
     s3_uri : string | S3Path
         the S3 URI of the cloud resource
     """
-    # create file system
-    fs = s3fs.S3FileSystem()
+    # create file system and get URL of file
+    try:
+        fs = s3fs.S3FileSystem(anon=True)
+        with fs.open(s3_uri, 'rb') as f:
+            return f.url()
+    except PermissionError:
+        # work-around for Roman Science Platform when acccessing private resources on the cloud
+        fs = s3fs.S3FileSystem()
+        with fs.open(s3_uri, 'rb') as f:
+            return f.url()
+
 
     # open resource and get URL
     with fs.open(s3_uri, 'rb') as f:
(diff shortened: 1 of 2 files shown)


astrocut/asdf_cutouts.py
```diff
@@ -25,8 +25,17 @@ def _get_cloud_http(s3_uri: Union[str, S3Path]) -> str:
...
+    # create file system and get URL of file
+    try:
+        fs = s3fs.S3FileSystem(anon=True)
+        with fs.open(s3_uri, 'rb') as f:
+            return f.url()
+    except PermissionError:
+        # work-around for Roman Science Platform when acccessing private resources on the cloud
+        fs = s3fs.S3FileSystem()
+        with fs.open(s3_uri, 'rb') as f:
+            return f.url()
```
havok2063, line 37: Is there a way to check for permissions usage, or access, outside of try-excepting for a `PermissionError`.  I'm wondering if there's a way to consolidate this check into a conditional for `anon`, e.g.
```
        fs = s3fs.S3FileSystem(anon=is_anonymous is True)
        with fs.open(s3_uri, 'rb') as f:
            return f.url()
```
If so, that might simplify this code a bit.  Otherwise, it's fine.  In that case, would we want to issue a log.warning within the `except` block to indicate what's happening?

snbianco: Hmm, I see your point! One option would be to send an HTTP request to the URL:
```
import requests
from urllib.parse import urlparse

uri = 's3://roman-sci-test-data-prod-summer-beta-test/ROMANISIM/DENSE_REGION/R0.5_DP0.5_PA0/r0000101001001001002_01101_0001_WFI18_cal.asdf'
parsed_uri = urlparse(uri)
url = f'https://{parsed_uri.netloc}.s3.amazonaws.com/{parsed_uri.path.lstrip('/')}'

response = requests.head(url)
if response.status_code == 200:
    print("Public")
elif response.status_code == 403:
    print("Private")
else:
    print("Else")
```
I wish this could be consolidated a bit more, but it would be a more robust check then hard-coding a bucket name. I'll add this and log messages to make the process more transparent.

astrocut/asdf_cutouts.py
```diff
@@ -16,19 +17,28 @@
...
+    # check if public or private by sending an HTTP request
+    s3_path = S3Path.from_uri(s3_uri) if isinstance(s3_uri, str) else s3_uri
+    url = f'https://{s3_path.bucket}.s3.amazonaws.com/{s3_path.key}'
+    resp = requests.head(url)
+    is_anon = False if resp.status_code == 403 else True
+    if verbose and not is_anon:
+        print(f'Attempting to access private S3 bucket: {s3_path.bucket}')
+
+    # create file system and get URL of file
+    fs = s3fs.S3FileSystem(anon=is_anon)
```
havok2063, line 41: This looks much cleaner to me and I like the approach here.  Does this add much overhead to the call?  I know `request.head` is a pretty lightweight request so I imagine it's negligible .  Do we need to add something in case this request fails or times out?  We could add a `timeout` [keyword](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts) to the request.

havok2063 (approved):
This looks good to me.  I had another question but modulo that I approve.

astrocut/asdf_cutouts.py
snbianco: Yep, I think `requests.head` is a good choice here since it doesn't download any body content. Adding a timeout is a good call as well, I will push another commit!

havok2063 (approved):

PR #124, opened 2024-07-30
Add parameters for S3 credentials

snbianco:
Added parameters to `asdf_cut` that allow the user to provide an S3 access key (`key`), secret access key (`secret`), or security token (`token`). These should only have an impact for private S3 buckets (`is_anon = False`), so I left in the logic to determine if a bucket is public or private.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -17,7 +17,8 @@ from astropy.coordinates import SkyCoord
 from astropy.modeling import models
 
 
-def _get_cloud_http(s3_uri: Union[str, S3Path], verbose: bool = False) -> str:
+def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = None,
+                    token: str = None, verbose: bool = False) -> str:
     """ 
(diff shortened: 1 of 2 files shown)


astrocut/asdf_cutouts.py
```diff
@@ -38,7 +45,7 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], verbose: bool = False) -> str:
         print(f'Attempting to access private S3 bucket: {s3_path.bucket}')
 
     # create file system and get URL of file
-    fs = s3fs.S3FileSystem(anon=is_anon)
+    fs = s3fs.S3FileSystem(anon=is_anon, key=key, secret=secret, token=token)
```
havok2063, line 48: Do we want a check here to make sure one of the access inputs is set when `is_anon` is False?

snbianco: For certain cases, a credential may not be needed. For instance, RSP doesn't need to pass credentials because the environment has direct access to the private S3 bucket.

havok2063: that makes sense, thanks!

havok2063 (approved):

PR #127, opened 2024-09-18
Add footprint finder code

snbianco:
This is a draft PR that adds the footprint finder code from TESScut. Leaving as draft since I haven't added tests yet, but I'd welcome any feedback on the actual implementation.

I tried to make things as general as possible in terms of functions and variable names, but some things are more TESS specific and will need to be pulled out into a wrapper later when we generalize. Namely, `_extract_sequence_information`, `_create_sequence_list`, and `_get_cube_files_from_sequence_obs` are more TESS-specific as of now. The same is true about certain parts of `cube_cut_from_footprint`, mainly variables.

Something I was unsure about is how to best handle multiprocessing. The `cube_cut_from_footprint` function takes a `threads` parameter to use as `max_workers` when using `cube_cut` on the gathered cube files. However, the `cube_cut` function also takes its own `threads` parameter. Should these be two separate parameters to `cube_cut_from_footprint`, or the same? Should `threads` in `cube_cut` just be set to `'auto'`?


astrocut/cube_cut_from_footprint.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+def _ffi_intersect(ffi_list: Table, polygon: SphericalPolygon):
+    """
+    Vectorizing the spherical_coordinate intersects_polygon function
+    """
+    def single_intersect(ffi, polygon):
+        return ffi.intersects_poly(polygon)
+
+    return np.vectorize(single_intersect)(ffi_list["polygon"], polygon)
```
snbianco, line 88: I tried added multiprocessing to this function, but found that there was no performance improvement (if anything, execution time went up because of thread overhead.) In any case, the `ra_dec_crossmatch` function takes < 1 sec to run.

scfleming:
My initial reaction is to say let's setup a single `n_threads` parameter for multi-threading of any function that can use it. While on paper it might be nice to say "use 8 threads for this one but 16 for that one", that sounds like over-engineering to me at this stage of the process, and it would be much simpler to have a single `n_threads` parameter used globally,

astrocut/cube_cut_from_footprint.py
falkben: when you say threads, do you actually mean threads or processes?

This is CPU bound, and currently, with the GIL, there will not be any improvement using threads. You would need to use multiple processes.

astrocut/cube_cut_from_footprint.py
```diff
@@ -0,0 +1,310 @@
...
+    Optional Parameters:
+        as_table: Return the footprint file as an Astropy Table
+        load_polys: Convert the s_region column to an array of SphericalPolygon objects
+    """
+    # Open footprint file with fsspec
+    # Use caching to help performance, but check that remote UID matches the local
+    # Expiry time is 1 week by default
+    with fsspec.open('filecache::' + s3_uri, s3={'anon': True}, 
+                     filecache={'cache_storage': 's3_cache', 'check_files': True}) as f:
+        ffis = json.loads(f.read())
```
falkben, line 71:
```suggestion
        ffis = json.load(f)
```

astrocut/cube_cut_from_footprint.py
snbianco: I did mean threads, but I just tried with `ProcessPoolExecutor`, and oddly enough, timing is a bit worse. It takes around 1 second longer with the process pool vs. vectorize. This is the code I'm using for comparison:

```
def single_intersect(ffi, polygon):
        return ffi.intersects_poly(polygon)


def _ffi_intersect_process(ffi_list: Table, polygon: SphericalPolygon, workers=None):
    """
    Vectorizing the spherical_coordinate intersects_polygon function
    """
    
    with ProcessPoolExecutor(max_workers=workers) as executor:
        # Map the function across the ffi_list["polygon"]
        results = list(executor.map(single_intersect, ffi_list["polygon"], [polygon] * len(ffi_list)))

    return np.array(results)
```

astrocut/cube_cut_from_footprint.py
```diff
@@ -0,0 +1,310 @@
...
+                verbose=verbose
+            )
+            return cutout
+        except Exception as e:
+            warnings.warn(f'Unable to generate cutout from {file_path}: {e}', AstropyWarning)
+            return None
+        
+    # Generate cutout from each cube file
+    cutout_files = []
+    if threads == 'auto' or threads > 1:
```
falkben, line 289: I am not sure threads here will help, and it may in fact hurt. `cube_cut` already uses threads for accessing the S3 data, so with this change, each cutout file would spawn that many threads. In my testing, there are diminishing returns after 8 threads. Since this could end up creating many times that number of threads, I expect we'd see thread contention here. 

If you set threads to 0, versus setting threads to "auto" or "8" what are the results here?

astrocut/__init__.py
```diff
@@ -28,3 +28,4 @@ class UnsupportedPythonError(Exception):
     from .cutout_processing import (path_to_footprints, center_on_path,  # noqa
                                     CutoutsCombiner, build_default_combine_function)  # noqa
     from .asdf_cutouts import asdf_cut, get_center_pixel  # noqa
+    from .cube_cut_from_footprint import cube_cut_from_footprint  # noqa
```
falkben, line 31: this hides the module, is that intended?

I think it would be better to name the module differently from the function that you are intending to export

astrocut/cube_cut_from_footprint.py
```diff
@@ -0,0 +1,310 @@
...
+        is a scalar number or a scalar `~astropy.units.Quantity`,
+        then a square cutout of ``cutout_size`` will be created.  If
+        ``cutout_size`` has two elements, they should be in ``(ny, nx)``
+        order.  Scalar numbers in ``cutout_size`` are assumed to be in
+        units of pixels. `~astropy.units.Quantity` objects must be in pixel or
+        angular units.
+    sequences : int, List[int], optional
+        Default None. Sequence(s) from which to generate cutouts. Can provide a single
+        sequence number as an int or a list of sequence numbers. If not specified, 
+        cutouts will be generated from all sequences that contain the cutout.
```
falkben, line 201: Maybe you were trying to generalize this but I'm not sure what sequences are. Are these sectors?

astrocut/cube_cut_from_footprint.py
falkben: I tried your attached code and couldn't get it to work in a different file without some modifications. I also am not sure it's correct, with the arguments you've specified in `map`.

However, I can confirm that using the process pool executor is no faster, and a tiny bit worse, no matter what number of processes I use. There is overhead to starting up subprocesses in Python, so perhaps that's the issue we're running into. Also, SphericalGeometry might have some multiprocessing itself, and so we're just getting in the way of that by starting multiple processes?

Running the below code reports the intersect taking just over 1 second on my machine, whereas the version in this PR takes around a second or just under a second.

```py
import functools
import sys
import time
from concurrent.futures import ProcessPoolExecutor
from multiprocessing import freeze_support

import numpy as np
from astropy.coordinates import SkyCoord
from astropy.table import Table
from spherical_geometry.polygon import SphericalPolygon

import astrocut.cube_cut_from_footprint

ccff = sys.modules["astrocut.cube_cut_from_footprint"]


def single_intersect(polygon, ffi_poly):
    return ffi_poly.intersects_poly(polygon)


def _ffi_intersect_process(ffi_list: Table, polygon: SphericalPolygon, workers=None):
    """
    Vectorizing the spherical_coordinate intersects_polygon function
    """

    start_t = time.monotonic()
    with ProcessPoolExecutor(max_workers=4) as executor:
        # Map the function across the ffi_list["polygon"]

        single_intersect_poly = functools.partial(single_intersect, polygon)
        results = executor.map(single_intersect_poly, ffi_list["polygon"], chunksize=64)
    print("took:", time.monotonic() - start_t)

    return np.array(list(results))


if __name__ == "__main__":
    # monkeypatch our version in
    ccff._ffi_intersect = _ffi_intersect_process

    footprint_file = "s3://tesscut-ops-footprints/tess_ffi_footprint_cache.json"

    all_ffis = ccff._get_s3_ffis(footprint_file, as_table=True, load_polys=True)

    coordinates = "217.42893801 -62.67949189"
    coordinates = SkyCoord(coordinates, unit="deg")

    cutout_size = ccff.parse_size_input(10)
    cone_results = ccff._ra_dec_crossmatch(all_ffis, coordinates, cutout_size, ccff.TESS_ARCSEC_PER_PX)

    print(cone_results)

```

astrocut/cube_cut_from_footprint.py
snbianco: Yes, they refer to sectors! I was trying to make the parameter more general and borrowed "sequence" from the CAOM field descriptions: https://mast.stsci.edu/api/v0/_c_a_o_mfields.html

astrocut/cube_cut_from_footprint.py
snbianco: Testing on my machine, I do see a performance improvement with a larger number of threads. It's more apparent when a sector isn't provided and more than 1 cutout is being generated. For example, these commands each generate 7 cutouts.

`cube_cut_from_footprint('130 30', cutout_size=50, output_dir='out_dir_2', threads=0)` --> 1 min, 23.4 sec
`cube_cut_from_footprint('130 30', cutout_size=50, output_dir='out_dir_2', threads=8)` --> 57.6 sec
`cube_cut_from_footprint('130 30', cutout_size=50, output_dir='out_dir_2', threads='auto')` --> 46.4 sec

astrocut/cube_cut_from_footprint.py
snbianco: I get a `TypeError` from this change because `json.loads()` won't accept a `BufferedReader` object.

astrocut/__init__.py
snbianco: Good point! I renamed the module to `footprint_cutouts`. 

I'm not sure if I know what you mean by hiding the module. You can still import the module on its own.

falkben: without the change to the name of the module, you couldn't easily import the module from another file. 

you had to resort to some trickery. e.g.:

```py
import astrocut.cube_cut_from_footprint

ccff = sys.modules["astrocut.cube_cut_from_footprint"]
```

falkben: note that the first import above doesn't import the module, it imported the function. but the module does get added to sys.modules when you import the function, so you can extract it from there

astrocut/cube_cut_from_footprint.py
falkben: i guess for a user, it might be a little unclear that for TESS this is sectors, and not cameras or ccds or anything like that. so maybe some documentation or examples would help here.

astrocut/cube_cut_from_footprint.py
falkben: is this before or after you made the change to use the same threads variable to pass into the `cube_cut` function?

astrocut/__init__.py
snbianco: I see, that makes sense!

astrocut/cube_cut_from_footprint.py
snbianco: Looks like I ran the test before, when `threads` was set to auto for `cube_cut`. When using the same `threads` variable, the call with `threads=0` takes a lot longer, which makes sense. I also see that there is less than a second difference between `threads=8` and `threads=auto`. 

Maybe it would be best to keep `threads` for `cube_cut` constant at some value, like 'auto' or 8? I think that using threads in `cube_cut_from_footprint` is still worthwhile for the performance improvement when making several cutouts at once, but performance does seem to stagnate after a certain point.

astrocut/cube_cut_from_footprint.py
snbianco: Added some more info and an example to the docstring in the latest commit! I'll also be updating the documentation at some point (probably next spring) and will definitely include examples there.

snbianco:
I added unit tests for the module, but there seems to be a problem with accessing the public footprint files on S3 in the runners. From what I can find online, this is a permissions issue. The odd thing is, we have other tests that access S3 resources and work fine.

`botocore.exceptions.ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden`

snbianco:
> I added unit tests for the module, but there seems to be a problem with accessing the public footprint files on S3 in the runners. From what I can find online, this is a permissions issue. The odd thing is, we have other tests that access S3 resources and work fine.
> 
> `botocore.exceptions.ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden`

I added a fixture to mock opening the footprint files with `fsspec`, and tests are passing now.

snbianco:
Marking as ready as tests and documentation have been added.

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,311 @@
...
+    # Generate cutout from each cube file
+    cutout_files = []
+    if threads == 'auto' or threads > 1:
+        # Parallel processing
+        if verbose:
+            print(f'Generating cutouts in parallel with max threads: {threads}')
+
+        # Use ThreadPoolExecutor for parallelization
+        max_workers = None if threads == "auto" else threads
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
```
falkben, line 297: trying this locally I'm not seeing a significant difference between "auto" and `0`. in fact, `0` is sometimes faster, though it is quite variable. I suggest removing this for now, as it adds additional complexity

This is what I was using for benchmarking:

```py
import time
from tempfile import TemporaryDirectory

from astrocut import cube_cut_from_footprint

coordinates = "217.42893801 -62.67949189"
with TemporaryDirectory() as tmp:
    start = time.monotonic()
    vals = cube_cut_from_footprint(coordinates, cutout_size=10, sequence=None, output_dir=tmp, threads=0)
    print("took:", time.monotonic() - start)
```

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,311 @@
...
+    # Get sector names and cube files that contain the cutout
+    cone_results = _ra_dec_crossmatch(all_ffis, coordinates, cutout_size, TESS_ARCSEC_PER_PX)
+    if not cone_results:
+        raise InvalidQueryError('The given coordinates were not found within the specified sequence(s).')
+    seq_list = _create_sequence_list(cone_results, product)
+    cube_files_mapping = _get_cube_files_from_sequence_obs(seq_list)
+    if verbose:
+        print(f'Found {len(cube_files_mapping)} matching cube files.')
+    base_file_path = "s3://stpubdata/tess/public/mast/" if product == 'SPOC' \
+        else "s3://stpubdata/tess/public/mast/tica/"
```
falkben, line 262: Are there some situations where someone might still want to make cutouts from a different storage path?

Could we make this an option?

One instance where someone might want a different option is if they have downloaded the cube, to make direct cutouts. But in that case, maybe they are just directly using `cube_cut`?

Another option might be if they've mounted stpubdata cube data onto their machine or cloud environment (w/ fuse or something else) in which case they'd rather use that path than the s3 path. E.g. TIKE cloud platform?

I do think we can probably default to these paths, though.

Or maybe we just indicate that this function is _only_ for making cutouts from s3 files? I guess it's already in the docstring, but it's not obvious from the module name or the function name.

falkben (commented):
There's an issue with accessing the footprint files in the bucket -- I have a ticket to work on that next week.

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,311 @@
...
+    Optional Parameters:
+        as_table: Return the footprint file as an Astropy Table
+        load_polys: Convert the s_region column to an array of SphericalPolygon objects
+    """
+    # Open footprint file with fsspec
+    # Use caching to help performance, but check that remote UID matches the local
+    # Expiry time is 1 week by default
+    s3_cache = os.path.join(os.path.dirname(os.path.abspath(__file__)), 's3_cache')
+    with fsspec.open('filecache::' + s3_uri, s3={'anon': True},
+                     filecache={'cache_storage': s3_cache, 'check_files': True}) as f:
```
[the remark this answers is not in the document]
snbianco: Here is the documentation on local caching in fsspec: https://filesystem-spec.readthedocs.io/en/latest/features.html#caching-files-locally

And here is the API for `CachingFileSystem` where the cache options are described: https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.implementations.cached.CachingFileSystem

It takes less than a second to fetch the file from S3, so an in-memory store like `TTLCache` is probably the way to go. This is all great to know!

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,311 @@
...
+    ...         output_dir='./cutouts')
+    ['./cutouts/tess-s0001-4-4/tess-s0001-4-4_83.406310_-62.489771_64x64_astrocut.fits',
+     './cutouts/tess-s0002-4-1/tess-s0002-4-1_83.406310_-62.489771_64x64_astrocut.fits']
+    """
+    
+    # Convert to SkyCoord
+    if not isinstance(coordinates, SkyCoord):
+        coordinates = SkyCoord(coordinates, unit='deg')
+    if verbose:
+        print('Coordinates: ', coordinates)
```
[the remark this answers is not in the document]
snbianco: Yep, switching to a logging framework is definitely a to-do.

astrocut/footprint_cutouts.py
snbianco: My thought was that users would use `cube_cut` in the case that they already have the path (whether local or cloud) to a single cube file. I think that providing a single file kind of defeats the purpose of the footprint lookup.

A mounted filesystem or a local path to many cube files is worth considering, but I do wonder how common that use case would be. Could we guarantee that the cube files match the footprints coming from S3?

I'm inclined to rename the function to something like `s3_cube_cut_from_footprint` and make a new issue to explore other options at a later time.

falkben: maybe `cloud_cube_cut_from_footprint` though it is a bit long?

though thinking more on it, i think what you have also works. i don't think there's a need to make an issue now. we can wait until a use case comes up.

snbianco: Do you think we could abbreviate and use `cloud_cube_cut_from_fp`? That may cause some confusion though since "fp" isn't too obvious.

falkben: between the two, i think i'd prefer just leaving off `cloud`. i'm not a big fan of acronym and `fp` isn't obvious

astrocut/footprint_cutouts.py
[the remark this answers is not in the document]
snbianco: Agreed! Created a JIRA ticket for this: https://jira.stsci.edu/browse/ASB-29048

falkben:
Since we're having a delay in opening up the cached footprint files on S3, another approach could be to download the footprint directly from CAOM through vo-tap interface on the first cutout.

This query gets the SPOC footprint table from CAOM and can be run from anywhere:

https://mast.stsci.edu/vo-tap/api/v0.1/caom/sync?FORMAT=json&LANG=ADQL&QUERY=SELECT+obs_id,+t_min,+t_max,+s_region,+target_name,+sequence_number+FROM+dbo.ObsPointing+WHERE+obs_collection=%27TESS%27+AND+dataproduct_type=%27image%27+AND+target_name=%27TESS+FFI%27

And this gets the TICA footprint (takes a bit longer since it's HLSP)

https://mast.stsci.edu/vo-tap/api/v0.1/caom/sync?FORMAT=json&LANG=ADQL&QUERY=SELECT%20obs_id,%20t_min,%20t_max,%20s_region,%20target_name,%20sequence_number%20FROM%20dbo.ObsPointing%20WHERE%20obs_collection=%27HLSP%27%20AND%20dataproduct_type=%27image%27%20AND%20target_name=%27TICA%20FFI%27

We manipulate that response in tesscut to create the footprint JSON file we store in S3 with a small bit of code but we could add that into astrocut as well.

Taking advantage of the cached footprint file in S3 is likely better long term, but we could use this method initially, for this PR

falkben:
Do we want to revisit any of the mocking now that we can access the footprints?

astrocut/tests/data/tess_ffi_footprints.json
falkben, line 1: do we still need this in the repo (and the one for TICA)?

snbianco: Good catch! Just removed those

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,325 @@
...
+    """
+    Vectorizing the spherical_coordinate intersects_polygon function
+    """
+    def single_intersect(ffi, polygon):
+        return ffi.intersects_poly(polygon)
+
+    return np.vectorize(single_intersect)(ffi_list['polygon'], polygon)
+
+
+def _ra_dec_crossmatch(all_ffis: Table, coord: SkyCoord, cutout_size, arcsec_per_px: int):
```
falkben, line 125: I think it might be useful to provide this function (or one that gets the footprint for you as well) as an external interface. I could see this being useful for people who want more control over the cube_cut.

Could be done in a separate PR or issue if you don't want to do it here.

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,325 @@
...
+from .utils.utils import parse_size_input
+
+TESS_ARCSEC_PER_PX = 21  # Number of arcseconds per pixel in a TESS image
+FFI_TTLCACHE = TTLCache(maxsize=10, ttl=900)  # Cache for FFI footprint files
+CUBE_CUT_THREADS = 8  # Number of threads to use in `cube_cut` function
+
+
+def _s_region_to_polygon(s_region: Column):
+    """
+    Takes in a s_region string of type POLYGON or CIRCLE and returns it as
```
falkben, line 34: i don't think this docstring is correct -- this function as currently written only supports `POLYGON`, not `CIRCLE`

astrocut/footprint_cutouts.py
```diff
@@ -0,0 +1,325 @@
...
+    with fsspec.open(s3_uri, s3={'anon': True}) as f:
+        ffis = json.load(f)
+
+    if load_polys:
+        ffis['polygon'] = _s_region_to_polygon(ffis['s_region'])
+
+    if as_table:
+        ffis = Table(ffis)
+
+    return ffis
```
falkben, line 112: Should this function be removed, for now?

falkben:
I was looking for the coverage results. 

Doesn't need to be handled in this PR, but it looks like there's a problem with codecov upload in the github action. From [Python 3.10 with numpy 1.23 and full coverage](https://github.com/spacetelescope/astrocut/actions/runs/11166073335/job/31039232647#logs) job:

```sh
[2024-10-03T16:46:10.843Z] ['info'] => Project root located at: /home/runner/work/astrocut/astrocut
[2024-10-03T16:46:10.846Z] ['info'] -> No token specified or token is empty
[2024-10-03T16:46:10.936Z] ['info'] Searching for coverage files...
[2024-10-03T16:46:10.973Z] ['info'] => Found 1 possible coverage files:
  ./coverage.xml
[2024-10-03T16:46:10.973Z] ['info'] Processing ./coverage.xml...
[2024-10-03T16:46:10.976Z] ['info'] Detected GitHub Actions as the CI provider.
[2024-10-03T16:46:11.306Z] ['info'] Pinging Codecov: https://codecov.io/upload/v4?package=github-action-2.1.0-uploader-0.8.0&token=*******&branch=footprint-finder&build=11166073335&build_url=https%3A%2F%2Fgithub.com%2Fspacetelescope%2Fastrocut%2Factions%2Fruns%2F11166073335&commit=9552cbb9116481980b67cbea56921b10ebb327db&job=CI&pr=127&service=github-actions&slug=spacetelescope%2Fastrocut&name=&tag=&flags=&parent=
[2024-10-03T16:46:11.478Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 366s."}

[2024-10-03T16:46:11.479Z] ['info'] Codecov will exit with status code 0. If you are expecting a non-zero exit code, please pass in the `-Z` flag
```

Anyways, coverage looks pretty good:

```
                                Stmts   Miss  Cover
---------------------------------------------------
astrocut/__init__.py               14      1    93%
astrocut/asdf_cutouts.py           82      3    96%
astrocut/cube_cut.py              388      4    99%
astrocut/cutout_processing.py     247     13    95%
astrocut/cutouts.py               244     26    89%
astrocut/exceptions.py             11      0   100%
astrocut/footprint_cutouts.py     139     11    92%
astrocut/make_cube.py             427     26    94%
astrocut/utils/__init__.py          0      0   100%
astrocut/utils/utils.py            86      4    95%
astrocut/utils/wcs_fitting.py      31      5    84%
---------------------------------------------------
                                 1669     93    94%
```

snbianco:
> I was looking for the coverage results.
> 
> Doesn't need to be handled in this PR, but it looks like there's a problem with codecov upload in the github action. From [Python 3.10 with numpy 1.23 and full coverage](https://github.com/spacetelescope/astrocut/actions/runs/11166073335/job/31039232647#logs) job:
> 
> ```shell
> [2024-10-03T16:46:10.843Z] ['info'] => Project root located at: /home/runner/work/astrocut/astrocut
> [2024-10-03T16:46:10.846Z] ['info'] -> No token specified or token is empty
> [2024-10-03T16:46:10.936Z] ['info'] Searching for coverage files...
> [2024-10-03T16:46:10.973Z] ['info'] => Found 1 possible coverage files:
>   ./coverage.xml
> [2024-10-03T16:46:10.973Z] ['info'] Processing ./coverage.xml...
> [2024-10-03T16:46:10.976Z] ['info'] Detected GitHub Actions as the CI provider.
> [2024-10-03T16:46:11.306Z] ['info'] Pinging Codecov: https://codecov.io/upload/v4?package=github-action-2.1.0-uploader-0.8.0&token=*******&branch=footprint-finder&build=11166073335&build_url=https%3A%2F%2Fgithub.com%2Fspacetelescope%2Fastrocut%2Factions%2Fruns%2F11166073335&commit=9552cbb9116481980b67cbea56921b10ebb327db&job=CI&pr=127&service=github-actions&slug=spacetelescope%2Fastrocut&name=&tag=&flags=&parent=
> [2024-10-03T16:46:11.478Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 366s."}
> 
> [2024-10-03T16:46:11.479Z] ['info'] Codecov will exit with status code 0. If you are expecting a non-zero exit code, please pass in the `-Z` flag
> ```
> 
> Anyways, coverage looks pretty good:
> 
> ```
>                                 Stmts   Miss  Cover
> ---------------------------------------------------
> astrocut/__init__.py               14      1    93%
> astrocut/asdf_cutouts.py           82      3    96%
> astrocut/cube_cut.py              388      4    99%
> astrocut/cutout_processing.py     247     13    95%
> astrocut/cutouts.py               244     26    89%
> astrocut/exceptions.py             11      0   100%
> astrocut/footprint_cutouts.py     139     11    92%
> astrocut/make_cube.py             427     26    94%
> astrocut/utils/__init__.py          0      0   100%
> astrocut/utils/utils.py            86      4    95%
> astrocut/utils/wcs_fitting.py      31      5    84%
> ---------------------------------------------------
>                                  1669     93    94%
> ```

Made an issue here: https://jira.stsci.edu/browse/ASB-29119

astrocut/footprint_cutouts.py
falkben: It looks like this docstring is still referring to CIRCLE but that doesn't look supported

snbianco: Are you looking at the most recent commit? I do see the change on that docstring.

falkben (approved):

astrocut/tests/test_utils.py
```diff
@@ -13,7 +16,51 @@
...
+    cutout_size = utils.parse_size_input((5, 10))
+    assert np.array_equal(cutout_size, np.array((5, 10)))
+
+    # With list as input
+    cutout_size = utils.parse_size_input([10, 5])
+    assert np.array_equal(cutout_size, np.array((10, 5)))
+
+    # With array as input
+    cutout_size = utils.parse_size_input(np.array((5, 10)))
+    assert np.array_equal(cutout_size, np.array((5, 10)))
```
havok2063, line 40: You can probably simplify this test by parametrizing over the inputs/outputs using `@pytest.mark.parametrize`

astrocut/tests/test_footprint_cutouts.py
```diff
@@ -0,0 +1,161 @@
...
+    
+    # Edge object that intersects
+    edge_intersect = SphericalPolygon.from_radec(lon=(340, 350, 350, 340),
+                                                 lat=(-15, -15, -5, -5),
+                                                 center=(345, -10))
+    
+    # Edge object that does not intersect
+    edge_nonintersect = SphericalPolygon.from_radec(lon=(340, 349, 349, 340),
+                                                    lat=(-15, -15, -5, -5),
+                                                    center=(345, -10))
```
havok2063, line 65: You could probably parametrize this test as well.

havok2063 (commented):
This all looks reasonable to me.  I only had a few small comments on some tests.

AlexReedy (dismissed):
Not much exposure to this codebase yet so can't comfortably comment on the details but this looks good on the overview to me. Ran the test suite last night (after changes) and it all came back good.

astrocut/tests/test_footprint_cutouts.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Create a table with this polygon
+    polygon_table = Table(names=['polygon'], dtype=[SphericalPolygon])
+    polygon_table['polygon'] = [polygon]
+
+    # Perform the intersection check
+    intersection = _ffi_intersect(polygon_table, cutout_sp)
+
+    # Assert the intersection result matches the expected value
+    assert intersection.value[0] == expected
```
falkben, line 63: maybe this assertion should be done with `is` instead?

```suggestion
    assert intersection.value[0] is expected
```

snbianco: Using `is` fails because `intersection.value[0]` is a `numpy.bool_` object.

falkben: oh, gotcha. okay!

falkben (approved):

PR #128, opened 2024-09-20
Documentation about odd vs. even cutout size in asdf_cut

snbianco:
Add documentation to warn users about cutout center accuracy when using an odd vs. even cutout size in `asdf_cut`. I added a note to the `cutout_size` parameter in the function docstring and updated the RST file with a warning block. I also changed the default cutout size for the function to be an odd number, and updated the docs to use odd sizes as well.


havok2063 (approved):
Seems reasonable to me.

PR #131, opened 2024-10-14
Add logging framework

snbianco:
Adds a logging framework. A few things to note:

1. The logger's default level is `INFO`, and most of the messages are printed to the `DEBUG` level.
2. Previously, we've been using a `verbose` parameter to log intermediate information. I didn't want to change the API by removing that parameter, so I wrote a function to set the log level based on the value of `verbose`. I'm conflicted on whether to deprecate the `verbose` parameters or leave them in for the sake of user-friendliness. As it is now, the `verbose` parameter overrides any level that the user sets the log to. Still, I think more users would know how to use `verbose` rather than set the log level themselves. Any thoughts are appreciated!
3. I restructured `__init__.py` to better separate logic.
4. I made a quick change to `test_asdf_cut.py` since one of the tests was leaving behind an output file.

### DIFF
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -45,7 +45,7 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = N
     resp = requests.head(url, timeout=10)
     is_anon = False if resp.status_code == 403 else True
     if not is_anon:
-        log.debug(f'Attempting to access private S3 bucket: {s3_path.bucket}')
+        log.info(f'Attempting to access private S3 bucket: {s3_path.bucket}')
 
     # create file system and get URL of file
     fs = s3fs.S3FileSystem(anon=is_anon, key=key, secret=secret, token=token)
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -304,7 +304,7 @@ def asdf_cut(input_file: Union[str, pathlib.Path, S3Path], ra: float, dec: float
     astropy.nddata.Cutout2D:
         An image cutout object.
     """
-
+    # Log messages based on verbosity
     _handle_verbose(verbose)
 
     # if file comes from AWS cloud bucket, get HTTP URL to open with asdf
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -45,7 +45,7 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = N
     resp = requests.head(url, timeout=10)
     is_anon = False if resp.status_code == 403 else True
     if not is_anon:
-        log.info(f'Attempting to access private S3 bucket: {s3_path.bucket}')
+        log.debug('Attempting to access private S3 bucket: %s', s3_path.bucket)
 
     # create file system and get URL of file
     fs = s3fs.S3FileSystem(anon=is_anon, key=key, secret=secret, token=token)
--- a/astrocut/asdf_cutouts.py
+++ b/astrocut/asdf_cutouts.py
@@ -16,6 +16,9 @@ from s3path import S3Path
 from astropy.coordinates import SkyCoord
 from astropy.modeling import models
 
+from . import log
+from .utils.utils import _handle_verbose
+
 
 def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = None,
                     token: str = None, verbose: bool = False) -> str:
@@ -41,8 +44,8 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = N
     url = f'https://{s3_path.bucket}.s3.amazonaws.com/{s3_path.key}'
     resp = requests.head(url, timeout=10)
     is_anon = False if resp.status_code == 403 else True
-    if verbose and not is_anon:
-        print(f'Attempting to access private S3 bucket: {s3_path.bucket}')
+    if not is_anon:
+        log.debug(f'Attempting to access private S3 bucket: {s3_path.bucket}')
 
     # create file system and get URL of file
     fs = s3fs.S3FileSystem(anon=is_anon, key=key, secret=secret, token=token)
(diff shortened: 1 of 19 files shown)


havok2063 (dismissed):
All your changes here seem reasonable to me.   I'm not sure if there were any print statements when errors occurred, but if so, `log.error` might be helpful to capture them and differentiate from info messages.    Also, are there any messages we'd like to display to the user even when verbosity is off?  If so, we could reshuffle the levels from `info/warning` to `debug/info`.  

I agree that `verbose` is a little more user-friendly than dealing with log levels.   If you wanted to retain a `verbose` flag but remove the boilerplate use in all the functions, one option could be to set up an astrocut user configuration file, like an `ini` or `yaml`, and move the `verbose` flag there.  This would allow `_handle_verbose` to be set once globally.   The upside is it's a convenient place to have all user config in one place.  A user could also set it and forget it.   The downside is it can be a bit clunkier for users to change the verbosity on the fly when running a function.   I'm not suggesting we need to make any changes here though, but it could be something to consider in the future, especially if we think the list of user configurable settings might grow.

astrocut/utils/utils.py
```diff
@@ -216,3 +217,7 @@ def get_fits(cutout_hdus, center_coord=None, output_path=None):
     return cutout_hdulist
 
 
+def _handle_verbose(verbose: bool):
+    """Set the log level according to the verbose parameter"""
+    level = logging.INFO if verbose else logging.WARNING
```
falkben, line 222: I think we should rework the levels here.

verbose should turn on DEBUG level.

and the default should probably include INFO

so anything that was always printed to the screen should be INFO or above. Anything that was gated by `verbose` should be debug.

falkben:
> I agree that `verbose` is a little more user-friendly than dealing with log levels. If you wanted to retain a `verbose` flag but remove the boilerplate use in all the functions, one option could be to set up an astrocut user configuration file, like an `ini` or `yaml`, and move the `verbose` flag there. This would allow `_handle_verbose` to be set once globally. The upside is it's a convenient place to have all user config in one place. A user could also set it and forget it. The downside is it can be a bit clunkier for users to change the verbosity on the fly when running a function. I'm not suggesting we need to make any changes here though, but it could be something to consider in the future, especially if we think the list of user configurable settings might grow.

astropy apparently has a config system already. i am not familiar with it, but perhaps we could also use it (or maybe `AstropyLogger` already does?)

https://docs.astropy.org/en/stable/config/index.html

astrocut/asdf_cutouts.py
```diff
@@ -41,8 +44,8 @@ def _get_cloud_http(s3_uri: Union[str, S3Path], key: str = None, secret: str = N
     url = f'https://{s3_path.bucket}.s3.amazonaws.com/{s3_path.key}'
     resp = requests.head(url, timeout=10)
     is_anon = False if resp.status_code == 403 else True
-    if verbose and not is_anon:
-        print(f'Attempting to access private S3 bucket: {s3_path.bucket}')
+    if not is_anon:
+        log.info(f'Attempting to access private S3 bucket: {s3_path.bucket}')
```
falkben, line 48: instead of formatting the string and passing it into the logger, as you have here with the fstring, and below with the `str.format()` calls, it is best practice to let the logger do the formatting. this is for security (in some places we do call astrocut as a web service, so some of these inputs may be untrusted and we don't want them logged onto a server), but there's another reason as well. If the logging level is skipped (e.g. Debug) the logger can avoid doing any str formatting if it's passed in this way, whereas if you format the string before passing it into the logger, you are always doing that (small) bit of work.

so in this case, my suggestion would be:

```suggestion
        log.info('Attempting to access private S3 bucket: %s', s3_path.bucket')
```

And should be applied to all the log formatting here.

astrocut/utils/utils.py
snbianco: Changed in newest commit!

snbianco:
I created a new ticket to set up a user configuration file here: https://jira.stsci.edu/browse/ASB-29237

havok2063 (approved):
Looks good to me!

astrocut/cube_cut.py
```diff
@@ -844,8 +843,9 @@ def cube_cut(
             if unsuccessful returns None.
         """
 
-        if verbose:
-            start_time = time()
+        start_time = time()
```
falkben, line 846: not a huge deal, but we should be using `time.monotonic()` or `time.perf_counter()` here instead.

https://docs.python.org/3/library/time.html#time.monotonic

astrocut/cube_cut.py
```diff
@@ -906,17 +902,15 @@ def cube_cut(
             # Get cutout wcs info
             cutout_wcs_full = self._get_full_cutout_wcs(cube[2].header)
             max_dist, sigma = self._fit_cutout_wcs(cutout_wcs_full, img_cutout.shape[1:])
-            if verbose:
-                print("Maximum distance between approximate and true location: {}".format(max_dist))
-                print("Error in approximate WCS (sigma): {}".format(sigma))
+            log.debug("Maximum distance between approximate and true location: %s", max_dist)
+            log.debug("Error in approximate WCS (sigma): %f", sigma)
```
falkben, line 906: When do we format with `%d` or `%f`?

astrocut/cutouts.py
```diff
@@ -8,7 +8,7 @@
 
 from time import time
 
-from astropy import log
+from astropy import log as astropy_log
```
falkben, line 11: should this still be using the astropy log?

astrocut/cube_cut.py
snbianco: I'm using %d for integers, %f for floats, and %s for other types. I added a precision to line 906 in the latest commit.

astrocut/cutouts.py
snbianco: Yes, I think because the `_hducut` function is using the logs from Astropy to determine whether or not a FITS file has SIP keywords in the WCS. I didn't write this code originally, but as far as I can tell from the Astropy source code, there isn't really a way around this without rerouting the logs: https://github.com/astropy/astropy/blob/main/astropy/wcs/wcs.py#L1299

falkben (approved):

AlexReedy (approved):
This looks good to me!

PR #132, opened 2024-11-11
Improve speed by using .section attribute rather than .data attribute

rlwastro:
Accessing the `.data` attribute on a compressed image causes the entire image to be decompressed and read into memory.  That defeats the ability of astrocut to reduce the IO requirements by reading only the necessary data.

Another benefit of the change is that it greatly reduces the memory requirements for the returned cutout.  The current version using the data section returns a subset of the full image array, which winds up dragging the entire array along with it.  The `.section` approach does not include the entire array.

This change instead uses the `.section` attribute to access the data, both when reading the pixels and when checking to see which extensions have images. The only remaining use of the `.data` attribute is for access to the cutout pixels. The implementation of the `section` attribute makes it efficient for accessing data in AWS S3 buckets as well.

The speedup ranges from a factor of 2 (for large cutouts) to at least a factor of 10 (for small cutouts).  

This change also fixes a bug in the application of the `memory_only` flag. When true, it should not write any output files to disk.  However, that was ignored when `single_outfile` is false.  This fixes that bug.  It was already working correctly when `single_outfile` is true.

### DIFF
--- a/astrocut/cutouts.py
+++ b/astrocut/cutouts.py
@@ -279,7 +279,7 @@ def fits_cut(input_files, coordinates, cutout_size, correct_wcs=False, extension
         with fits.open(in_fle, mode='denywrite', memmap=True, fsspec_kwargs=fsspec_kwargs) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.size > 0) for x in hdulist])[0]
+            all_inds = np.where([hdu.is_image and hdu.size > 0 for hdu in hdulist])[0]
             cutout_inds = _parse_extensions(all_inds, in_fle, extension)
 
             num_cutouts += len(cutout_inds)
@@ -537,7 +537,7 @@ def img_cut(input_files, coordinates, cutout_size, stretch='asinh', minmax_perce
         with fits.open(in_fle, mode='denywrite', memmap=True) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.size > 0) for x in hdulist])[0]
+            all_inds = np.where([hdu.is_image and hdu.size > 0 for hdu in hdulist])[0]
             cutout_inds = _parse_extensions(all_inds, in_fle, extension)
 
             for ind in cutout_inds:   
--- a/astrocut/cutouts.py
+++ b/astrocut/cutouts.py
@@ -77,7 +77,7 @@ def _hducut(img_hdu, center_coord, cutout_size, correct_wcs=False, verbose=False
             for log_rec in log_list:
                 astropy_log.log(log_rec.levelno, log_rec.msg, extra={"origin": log_rec.name})
 
-    img_data = img_hdu.data
+    img_data = img_hdu.section
 
     log.debug("Original image shape: %s", img_data.shape)
 
@@ -112,7 +112,7 @@ def _hducut(img_hdu, center_coord, cutout_size, correct_wcs=False, verbose=False
         padding[0, 1] = ymax - ymax_img
         ymax = ymax_img  
         
-    img_cutout = img_hdu.data[ymin:ymax, xmin:xmax]
+    img_cutout = img_hdu.section[ymin:ymax, xmin:xmax]
 
     # Adding padding to the cutout so that it's the expected size
     if padding.any():  # only do if we need to pad
@@ -279,7 +279,7 @@ def fits_cut(input_files, coordinates, cutout_size, correct_wcs=False, extension
         with fits.open(in_fle, mode='denywrite', memmap=True, fsspec_kwargs=fsspec_kwargs) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.data is not None) for x in hdulist])[0]
+            all_inds = np.where([x.is_image and (x.section.shape != ()) for x in hdulist])[0]
             cutout_inds = _parse_extensions(all_inds, in_fle, extension)
 
             num_cutouts += len(cutout_inds)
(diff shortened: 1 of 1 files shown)


rlwastro:
@snbianco @falkben Just tagging you to keep this fix on your radar.  I added a bit of text to the description.  I only recently realized this is not just an impact on speed.  Cutouts returned in memory also apparently return a view into the full (uncompressed) array, meaning the memory usage of the underlying data can be much larger than the cutout.  I have an example that extracts 250 cutouts of size 256x256 pixels (total size about 65 MB).  The resulting list from the current code was going to take 39 GB (!), which overflowed the 15 GB memory limit on the machine I was using.  The new code using `.section` works correctly and avoids the memory issues.

I think we should try to get this merged soon as possible since it has some serious impacts on performance.

astrocut/cutouts.py
```diff
@@ -279,7 +279,7 @@ def fits_cut(input_files, coordinates, cutout_size, correct_wcs=False, extension
         with fits.open(in_fle, mode='denywrite', memmap=True, fsspec_kwargs=fsspec_kwargs) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.data is not None) for x in hdulist])[0]
+            all_inds = np.where([x.is_image and (x.section.shape != ()) for x in hdulist])[0]
```
snbianco, line 282: For this line and 540, what was the rationale for changing this comparison? Could we use `x.section is not None` or `not x.size`?

snbianco (commented):
Hi Rick! Sorry it's taken me a while to get to this - I've been at ADASS this past week. Thank you for this PR; I'm excited about these changes and this was a really great catch. I'm a little concerned that we never noticed the bug with the `memory_only` flag, so I'll probably make another PR to add test cases sometime next week.

I had one minor comment/question, but these changes look good to me.

astrocut/cutouts.py
rlwastro: Hi Sam! This is the line I wondered about most myself. I did some testing with `.section` and found that `x.section` is not None, even for empty primary HDUs.  So I know a simple `x.section is not None` won't work.  The `x.section.shape != ()` weeds out the empty primary HDU.  But `x.size != 0` would also eliminate the empty primary HDU.

It is necessary to include the `x.is_image` test to avoid selecting table HDUs.  I tested a file with a BinTableHDU and found that (1) `x.size` is not zero, (2) `x.is_image` is False, and (3) `x.section` is not defined (raises an AttributeError).  

So that's how I came up with my code. I think either of these could work:
```python
x.is_image and (x.section.shape != ())
```
or
```python
x.is_image and (x.size > 0)
```
I do kinda like the `x.size` version of that better.  So I'd be happy with that change.

rlwastro: I just made that change, see what you think.

astrocut/cutouts.py
```diff
@@ -279,7 +279,7 @@ def fits_cut(input_files, coordinates, cutout_size, correct_wcs=False, extension
         with fits.open(in_fle, mode='denywrite', memmap=True, fsspec_kwargs=fsspec_kwargs) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.data is not None) for x in hdulist])[0]
+            all_inds = np.where([x.is_image and (x.size > 0) for x in hdulist])[0]
```
snbianco, line 282:
```suggestion
            all_inds = np.where([hdu.is_image and hdu.size > 0 for hdu in hdulist])[0]
```

astrocut/cutouts.py
```diff
@@ -533,7 +537,7 @@ def img_cut(input_files, coordinates, cutout_size, stretch='asinh', minmax_perce
         with fits.open(in_fle, mode='denywrite', memmap=True) as hdulist:
 
             # Sorting out which extension(s) to cutout
-            all_inds = np.where([x.is_image and (x.data is not None) for x in hdulist])[0]
+            all_inds = np.where([x.is_image and (x.size > 0) for x in hdulist])[0]
```
snbianco, line 540:
```suggestion
            all_inds = np.where([hdu.is_image and hdu.size > 0 for hdu in hdulist])[0]
```

astrocut/cutouts.py
rlwastro: Looks good to me - I committed those changes.

snbianco (approved):
Looks good to me! I'll make another PR with updated tests to address the coverage issue. Thanks again for this PR, this is a huge help!

falkben:
Very nice. Sorry it took a while for me to take a look at this. I wonder if we need similar change (`.data` -> `.section`) for this part of code:

https://github.com/spacetelescope/astrocut/blob/main/astrocut/cutout_processing.py#L463

and maybe that would speed up the moving target cutouts considerably?

rlwastro:
@falkben That's a good question.  I think the answer is yes, we should make the same change there.  If this code that pieces together images from different skycells gets applied to Pan-STARRS images (which is on my list of things to work on), it will be much faster if `.data` is changed to `.section` in that code.

I believe these changes will not make much difference in speed for uncompressed images (e.g., TESS data cubes and HAP-MVM images).  But I have not actually done timing tests on uncompressed images in the S3 bucket to see whether it has an effect on those images too.  And it is possible that the `.section` approach will improve memory handling even for uncompressed images.  I think it can't hurt in any case to make the change you suggest.

rlwastro:
@snbianco I have not quite been able to figure out what the parameters are for `build_default_define_function`. But assuming that they are already image cutouts that have been extracted from the larger images, I think you are correct. And in fact that line of code:
```python
img_arrs = np.array([hdu.data for hdu in template_hdu_arr])
```
... only really makes sense if the `data` elements are used rather than the `section` element.  It implies that we are stacking up all the individual cutouts into a big cube.

So yes, I agree with you!

PR #134, opened 2024-11-25
Better tests for writing to memory only

snbianco:
Modified unit tests after discovery of bug in #132. Now, tests check that no files are written if the `memory_only` flag is set to `True` for both a single output file and multiple output files.

Also, only makes a new directory for output if `memory_only` is `False`.


havok2063 (approved):
Looks good.

PR #142, opened 2025-02-17
use stpubdata for tesscut footprints

falkben:


astrocut/footprint_cutouts.py
```diff
@@ -51,9 +52,38 @@ def ind_sregion_to_polygon(s_reg):
...
+    ----------
+    as_table : bool, optional
+        Default False. Return the footprint file as an Astropy Table
+    load_polys : bool, optional
+        Default False. Convert the s_region column to an array of SphericalPolygon objects
+    """
+    # Open footprint file with fsspec
+    open_file_container = fsspec.open(s3_uri, s3={'anon': True})
+    with open_file_container as f:
+        ffis = json.load(f)
```
snbianco, line 73: Can we simplify this to:

```suggestion
    with fsspec.open(s3_uri, s3={'anon':True}) as f:
        ffis = json.load(f)
```

snbianco (commented):
Thank you for this PR! I had one minor comment, but otherwise looks good.

falkben:
Does this need an update to the changelog?

snbianco:
> Does this need an update to the changelog?

No, it's outdated. I'll make another PR to update it, and it's in my backlog to create a GitHub Action to check that an entry has been added for each PR.

snbianco (approved):

PR #143, opened 2025-02-19
Generalize CubeFactory classes

snbianco:
Generalizing the cube factory classes so that `TicaCubeFactory` is a child of `CubeFactory`.

A few considerations:

- This is not completely "general" in that the parent `CubeFactory` class is built for SPOC FFIs. Unfortunately, we can't do much about this without changing the existing API entirely. I did my best to factor all of the TESS-specific content into variables and methods that can be easily overridden in future child classes.
- The tests were kept largely the same, but I did add some assertions and restructured things a bit.
- There was some code in `_update_info_table` that was opening FITS files within a loop. I fixed this to only open the files once.
- I moved the logic for `update_cube` into `CubeFactory` (when it was originally just in `TicaCubeFactory`) and renamed the method to be public.


astrocut/CubeFactory.py
```diff
@@ -0,0 +1,670 @@
...
+from .exceptions import DataWarning, InvalidInputError
+from .utils.utils import _handle_verbose
+
+
+class CubeFactory():
+    """
+    Class for creating image cubes. This class is built to accept TESS SPOC FFI files,
+    but can be extended to work with other types of image files.
+    
+    Methods
```
havok2063, line 32: Maybe this is already the plan, but we might want to make this an abstract class, and make a new `SpocCubeFactory`, that, along with `TicaCubeFactory` could subclass from.

havok2063 (approved):
This all looks good to me.  It's mostly re-shuffling into classes.   I think it's ok to make large breaking changes in service of our larger generalization roadmap.  We can always issue a DeprecationWarning and make a major version bump to indicate API changes. 

For the cube stuff, I'd be curious to think about a good framework for easily slotting in JWST and SDSS MaNGA cubes, or potentially HLSP cubes.  It's a little tricky with the different z axis but it's kind of similar.

scfleming:
> This all looks good to me. It's mostly re-shuffling into classes. I think it's ok to make large breaking changes in service of our larger generalization roadmap. We can always issue a DeprecationWarning and make a major version bump to indicate API changes.

Breaking TESS cutouts from cubes, specifically, is NOT an inexpensive breaking change.  Other facets of astrocut I tend to agree with you on this philosophy.  Introducing backwards incompatibility with TESS cutouts from cubes should be an absolute last resort, and done with lots of planning.

snbianco:
My inclination was to leave the SPOC-specific logic in `CubeFactory` to avoid the large breaking change. I suppose we could make an abstract parent class and call it something else, and then have `CubeFactory` and `TicaCubeFactory` inherit from that. In the interest of time though, and since it won't have any effect on the API, I would probably wait to implement that in another PR.

PR #146, opened 2025-02-25
Cube Cutout Generalization

snbianco:
More generalization stuff!!! Only one more to go after this 🎉 

A few notes:
- Since the current implementation is so TESS-specific, I chose to make a concrete class, `TessCubeCutout` that inherits from abstract `CubeCutout`. I think I moved all the TESS-specific logic into the child class, but I may have missed some pieces.
- I followed the pattern set by the image cutout classes and included an inner class, `CubeCutoutInstance`, that has helpful attributes for each singular cutout.
- You'll notice that I changed around some of the errors on the other classes. I think I've settled on using `InvalidInputError` for invalid inputs that require little to no processing (not a valid option, not the right data type, etc.). `InvalidQueryError` will be for calls that do some processing on "valid" inputs, but ultimately fail due to some issue (no data, out of footprint, etc.)


astrocut/CubeCutout.py
```diff
@@ -0,0 +1,564 @@
...
+                        header[kwd] = (val, primary_header.comments[kwd])
+
+    @abstractmethod
+    def _build_tpf(self, cube_fits: fits.HDUList, cutout: 'CubeCutoutInstance') -> fits.HDUList:
+        """
+        Build the cutout target pixel file (TPF).
+
+        This method is abstract and should be defined in subclasses.
+        """
+        pass
```
havok2063, line 273: Will `CubeCutout` be used as the parent for any non-TESS cube cutout classes, e.g. JWST or MaNGA cubes?  If so,  maybe this should be moved to `TessCubeCutout`, or moved into some intermediate `TessBase` class if it needs to be an `abstractmethod`.

astrocut/cube_cut.py
```diff
@@ -38,754 +19,6 @@ class CutoutFactory():
...
-        
-        aperture_hdu.header['INHERIT'] = True
-    
-        cutout_hdu_list = fits.HDUList([primary_hdu, table_hdu, aperture_hdu])
-        
-        self._apply_header_inherit(cutout_hdu_list)
-
-        return cutout_hdu_list
-
     def cube_cut(
```
havok2063, line 22: To mirror `fits_cut` and `img_cut`, should this be made into a convenience function?

havok2063 (dismissed):
Overall looks good to me.

astrocut/cube_cut.py
snbianco: Do you mean to take it out of `CubeFactory` or add another function entirely? I chose not to take it out of `CubeFactory` because it wouldn't be backwards compatible.

snbianco: I like this idea! New commit adds a stand-alone `cube_cut` function.

snbianco:
Sorry, I messed up the rebase 😭 This should be the last commit if the CI passes.

havok2063 (approved):

PR #149, opened 2025-03-11
Footprint Cutouts Generalization

snbianco:
Last implementation PR for generalization!
 
Some notes:
* This class largely relies on `TessCubeCutout` to do the actual cutouts, so the majority of new logic is in fetching the footprint file and crossmatching to find matching cubes.
* `FootprintCutout` has more general logic while `TessFootprintCutout` has TESS-specific logic (akin to `CubeCutout` and `TessCubeCutout`)
* I made some changes here to fix documentation errors/warnings.


astrocut/CubeCutout.py
```diff
@@ -163,6 +163,8 @@ def _parse_table_info(self, table_data: fits.fitsrec.FITS_rec) -> WCS:
 
             # Making sure we have a row with wcs info.
             row = table_data[data_ind]
+            print('hello', self._wcs_axes_keyword)
```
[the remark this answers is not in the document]
snbianco: Yep, that's my bad! I made the change and forgot to push it.

tox.ini
```diff
@@ -70,7 +70,9 @@ commands =
 changedir = docs
 description = invoke sphinx-build to build the HTML docs
 extras = docs
-deps = sphinx_rtd_theme
+deps = 
+    sphinx_rtd_theme
+    matplotlib
```
falkben, line 75: what do we use matplotlib for?

astrocut/tests/test_TessFootprintCutout.py
```diff
@@ -0,0 +1,220 @@
...
+    return 5
+
+
+@pytest.fixture
+def coordinates():
+    """Fixture to return the coordinates at the center of the images"""
+    return SkyCoord('350 -80', unit='deg')
+
+
+def check_output_tpf(tpf, ffi_type, sequences=[], cutout_size=5):
```
falkben, line 29: should this use the cutout_size fixture above?

astrocut/tests/test_TessFootprintCutout.py
falkben, line 1: did the tests change here from what was in `test_footprint_cutouts.py`?

tox.ini
snbianco: Sphinx uses `plot_directive` from matplotlib. Without it, I get a warning when trying to build the docs:
```
/Users/sbianco/astrocut/.tox/build_docs/lib/python3.11/site-packages/sphinx_astropy/conf/v1.py:221: UserWarning: matplotlib's plot_directive could not be imported. Inline plots will not be included in the output
```

astrocut/tests/test_TessFootprintCutout.py
snbianco: Several of them are largely the same, but they use the new class instead of the `cube_cut_from_footprint` convenience function. I removed a few that were testing specific functions since those are now non-static, private class functions. Because the class is structured a bit differently, there's an additional test for writing to a file and additional assertions that check expected class attributes. The last test, `test_cube_cut_from_footprint` is a test for the convenience method.

astrocut/tests/test_TessFootprintCutout.py
snbianco: Since this is a helper function and not an actual test function, I don't think it can.

astrocut/FootprintCutout.py
```diff
@@ -0,0 +1,257 @@
...
+        # Assigning other attributes
+        if isinstance(sequence, int):
+            sequence = [sequence]  # Convert to list
+        self._sequence = sequence
+
+        # Populate these in child classes 
+        self._s3_footprint_cache = None  # S3 URI to footprint cache file
+        self._arcsec_per_px = None  # Number of arcseconds per pixel in an image
+    
+    @cached(cache=FFI_TTLCACHE, lock=Lock())
```
falkben, line 71: if you use `lru_cache` or `cache` on an instance method such as this

> instances of the class encapsulating that method never get garbage collected within the lifetime of the process holding them.

I did a tiny bit of digging and it looks like functools.lru_cache is not directly used by cachetools, but it might still be subject to the same problems and I'm also not 100% confident I was able to search fully, but I suspect that it will result in the same issues. If you like, I can double check, in a toy example.

http://rednafi.com/python/lru_cache_on_methods/

Anthony Sottile made a video about this as well, in case it's helpful. it's been a while since i watched it: https://www.youtube.com/watch?v=sVjtp6tGo0g

Anthony shows towards the end of the video how to get around this while still keeping it as a class method, but I think it might make sense to keep this function outside of the class itself since you might want to use this function without making a cutout anyways.

astrocut/tests/test_TessFootprintCutout.py
falkben: I missed that -- personally don't care too much for helper functions and prefer checks to be inlined in the tests since it makes it much easier to debug. 

I'm also a little curious why we are using a fixture instead of a global/constant, but I digress, this is just style and not really important.

tox.ini
falkben: can we make a comment? i find figuring out what is going on in a tox.ini already pretty complicated, so any help there to demystify things would be greatly appreciated.

astrocut/FootprintCutout.py
snbianco: This is really good to know! Thank you for looking into it, and for the video. I wrote up a quick example with cachetools and it does seem to have the same issue. I agree that it would be better to move the function outside the class.

astrocut/FootprintCutout.py
```diff
@@ -39,8 +39,6 @@ class FootprintCutout(Cutout, ABC):
 
     Methods
     -------
-    _get_ffis()
-        Fetches footprints for Full Frame Images (FFIs) from the S3 cloud.
     _get_files_from_cone_results(cone_results)
         Returns the files that contain the cutout.
     cutout()
```
falkben, line 44: I wonder if these docstrings should contain these private methods?

astrocut/footprint_cutouts.py
```diff
@@ -253,6 +70,8 @@ def cube_cut_from_footprint(coordinates: Union[str, SkyCoord], cutout_size,
         For the TESS mission, this parameter corresponds to sectors.
     product : str, optional
         Default 'SPOC'. The product type to make the cutouts from.
+    memory_only : bool, optional
+        Default False. If True, the cutouts are stored in memory and not written to disk.
     output_dir : str, optional
         Default '.'. The path to which output files are saved.
         The current directory is default.
```
falkben, line 77: commenting here but issue is on line 83. docstring for what this returns needs to be updated in the case of memory_only

personal bias, but i am not a huge proponent of docstrings in the first place, as docstrings get out of date, no real way to verify that they are correct.

Instead I prefer type annotations, because static type checking can verify, and editors (users) shows what functions take as arguments and what they return. 

Either way, i think it would be good to add type annotations of the return values for any user facing functions.

falkben:
is there any documentation that needs to be updated?

astrocut/footprint_cutouts.py
falkben: yes, i agree and perhaps I should have phrased my comment differently.

If from the name and typing of the argument it's not clear, the docstring is a good place to put it.

astrocut/FootprintCutout.py
```diff
@@ -0,0 +1,263 @@
...
+    
+    @abstractmethod
+    def _get_files_from_cone_results(self, cone_results: Table) -> dict:
+        """
+        Converts a `~astropy.table.Table` of cone search results to a list of dictionaries containing
+        metadata for each cloud file that intersects with the cutout.
+        
+        This method is abstract and should be implemented in subclasses.
+        """
+        raise NotImplementedError('Subclasses must implement this method.')
```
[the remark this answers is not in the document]
falkben: nice, i found the docs around that a bit confusing. agreed that this didn't need to change, but also it doesn't need to change back.

snbianco:
> is there any documentation that needs to be updated?

I actually made a different PR for all of the documentation updates here: https://github.com/spacetelescope/astrocut/pull/150

I was waiting to move it into review until all of the implementation PRs were merged.

astrocut/FootprintCutout.py
snbianco: Latest commit removes private methods from class docstrings.

astrocut/footprint_cutouts.py
snbianco: The latest commit adds type annotations for the return values of the functions defined in `footprint_cutouts.py`.

snbianco:
Is this PR okay to merge? I'd like to have next week to get through the documentation updates.

astrocut/FootprintCutout.py
falkben, line 1: I see that we have this pattern in a lot of files in this repo, but modules (filenames) should be all lowercase for PEP8 (https://peps.python.org/pep-0008/#package-and-module-names). so this should be `footprint_cutout.py`

but we already have `footprint_cutouts.py` (plural) ...

I think this might also be confusing for users. should we rename this, or just include it in the existing `footprint_cutouts.py` module? i'm not sure what makes the most sense here

BTW, I wondered if astropy and astroquery follow this PEP8 convention, and of the 880 .py files in astropy, none have uppercase letters. In astroquery (513 files), there are a few (10), I'm not sure why, and their history, mostly tests or utils, though, not user facing I think.

```sh
$ find . -name '*.py' | grep "[A-Z].*"        
./utils/tap/xmlparser/tableSaxParser.py
./utils/tap/xmlparser/jobListSaxParser.py
./utils/tap/xmlparser/sharedItemsSaxParser.py
./utils/tap/xmlparser/groupSaxParser.py
./utils/tap/xmlparser/jobSaxParser.py
./utils/tap/conn/tests/DummyConn.py
./utils/tap/conn/tests/DummyConnHandler.py
./utils/tap/conn/tests/DummyResponse.py
./esa/euclid/tests/DummyTapHandler.py
./esa/jwst/tests/DummyTapHandler.py
```

snbianco: This is a great point. I would prefer that the files that include the classes match the name of the class, so I'm leaning towards naming the file `footprint_cutout.py` and moving the convenience functions into that file.

falkben: it looks like CI is not running. There are also some merge conflicts, I think.

snbianco: Oops, I missed that! Should be good now.

falkben (approved):

PR #150, opened 2025-03-24
Generalization docs

snbianco:
Final PR for generalization! This encompasses documentation changes and a few other minor edits.

One thing that I want to highlight for feedback are the API pages. [Here's one example](https://astrocut--150.org.readthedocs.build/en/150/api/astrocut.ASDFCutout.html#astrocut.ASDFCutout).

They are automatically generated by [automodapi](https://sphinx-automodapi.readthedocs.io/en/latest/), and there is a lot of repeated information. For example, we have the "Attributes" and "Methods" documentation that gets taken from the docstring. Then, automodapi automatically creates other summary and documentation sections. I see a couple paths to take here:

1. Remove the class docstrings and use only the automatic documentation from automodapi. 
    - Pro: Simpler, includes inherited methods and attributes.
    - Con: Seems to only pick up attributes with a `@property` tag.
2. Abandon automodapi entirely and use a different tool.
    - Pro: Find something with more customization options.
    - Con: We are limited on time to implement something like this.
3. Leave the documentation as is.
    - Pro: I'm personally a fan of class docstrings, so I would like to keep them in if possible. I checked to see if automodapi has a configuration setting to ignore class docstrings, but it doesn't seem to.
    - Con: Is redundant and may confuse users.

I would be interested in checking out any documentation tools that are already in use on other MAST projects.


havok2063:
My vote would be for 3.  I also find code docstrings pretty useful.  I'd prefer to have some docs in the codebase so as not to have to go do the web docs all the time.

snbianco:
Pinging this PR again since I'd like to make the release before next Friday.

PR #153, opened 2025-04-08
Fix for ASDF Storage Bug

### LINKED ISSUE #152: ASDF cutout storage size equivalent to non-cutout storage size.
When using `asdf_cut` cut and outputting an ASDF file, the storage space required by the file is the size of the non-cutout array. The FITS file output has the correct size. For example, in the below code (using simulated Roman data), if I use `asdf_cut` to export to a FITS file, the storage usage ends up being ~ 160KB, which is expected for the resulting 200x200 (32-bit float) cutout requested. If I export to a ASDF file, the resulting storage is ~67MB, which is the size expected for the 4088x4088 original image the cutout is being taken from.

I inspected the object memory usage via [Pympler](https://pympler.readthedocs.io/en/latest/), and found that data array in the ASDF tree when loading the ASDF file is indeed where the ~67MB of memory is being used. If I force copy this data to a NumPy array, it immediately drops down to the expected ~ 160KB. Further, I inspected the data array object in the ASDF tree and found its internal `_offset` and `_striding` values to be strange (at `31812420` and `[16352, 4]` respectively). This suggests that the internal memory layout is still set up for the original 4088x4088 image it was taken from, and so the array is still taking up this amount of space in memory, despite having a size of 200x200. One potential fix for this would be to force a copy of the data array to a new NumPy array before copying it into the the ASDF tree to force it to restructure the memory layout correctly.

```python
from copy import deepcopy
from pathlib import Path

import asdf
import numpy as np
import roman_datamodels as rdm
from astrocut import asdf_cut

asdf.get_config().validate_on_read = False
asdf_file_path = Path('r0003201001001001004_0001_wfi11_f106_cal.asdf')

with asdf.open(asdf_file_path) as af:
    dm = rdm.open(af)
    data = deepcopy(dm.data)
    gwcs = deepcopy(dm.meta.wcs)

cutout_size = 200
pix_x, pix_y = np.shape(data)[0] // 2, np.shape(data)[1] // 2
center_coord = gwcs(pix_x, pix_y, with_units=True)
coord = center_coord

asdf_cut(asdf_file_path, coord.ra, coord.dec, cutout_size=cutout_size, output_file='example_output.fits')
asdf_cut(asdf_file_path, coord.ra, coord.dec, cutout_size=cutout_size, output_file='example_output.asdf')
```

This is on `astrocut==0.12.0`.

snbianco:
Fixes #152 

Previously, we were not accounting for the fact that `Cutout2D` returns a view into the original data and not a new array. This PR copies `cutout.data` into a new numpy array so that the memory layout restructures itself.


havok2063 (approved):
looks good

PR #161, opened 2025-08-05
Remove from footprint cutouts, deprecate product parameter

snbianco:


falkben:
For tesscut, I was thinking of just leaving product as a query argument in but setting it to a default SPOC. If we're deprecating here, I could try adding deprecation in tesscut as well. What do you think?

falkben:
Should this still be draft?

snbianco:
Changelog updated!

snbianco:
@falkben Is this PR good to merge?

PR #162, opened 2025-08-14
Remove TICA fully

snbianco:
Fully removing TICA cubes and cutouts from Astrocut after we deprecated in 1.1.0.


snbianco:
Actually, since so many of these changes are style, I'm gonna leave this in draft and make another PR to make all the style changes at once.

PR #164, opened 2025-09-25
Changelog check

snbianco:
Adding a CI job to check for a changelog entry on pull requests. You can set the `no-changelog-entry-needed` label to override this check.


snbianco:
I chose this particular one because Astroquery uses it, and it was developed by the Scientific Python Project. I think I would want to stay away from an automatic action that updates based on commit history since commit messages aren't always so descriptive (or readable). Some commits would also be internal changes that probably shouldn't be exposed to users.

PR #177, opened 2026-03-04
Allow user to pass in fsspec_kwargs to FITSCutout

dsarmiento:
Add the `fsspec_kwargs` argument to `FITSCutout` constructor. This will allow the user to pass in custom fsspec settings during the opening of a FITS file.

On the cloud we saw dramatic improvement when setting this to `1MB` as there is less data to load for a 256x256 PS1 cutout compared to the package default of `50MB`.

It reduced our cutout time on AWS ECS from 2s to about 650ms.

### DIFF
--- a/astrocut/fits_cutout.py
+++ b/astrocut/fits_cutout.py
@@ -698,7 +698,8 @@ def img_cut(input_files: List[Union[str, Path, S3Path]], coordinates: Union[SkyC
             invert: bool = False, img_format: str = '.jpg', colorize: bool = False,
             cutout_prefix: str = 'cutout', output_dir: Union[str, Path] = '.',
             extension: Optional[Union[int, List[int], Literal['all']]] = None, fill_value: Union[int, float] = np.nan,
-            limit_rounding_method: str = 'round', verbose=False, fsspec_kwargs: Optional[dict] = None) -> Union[str, List[str]]:
+            limit_rounding_method: str = 'round', verbose=False, fsspec_kwargs: Optional[dict] = None
+            ) -> Union[str, List[str]]:
     """
     Takes one or more fits files with the same WCS/pointing, makes the same cutout in each file,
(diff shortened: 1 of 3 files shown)


astrocut/fits_cutout.py
```diff
@@ -59,15 +59,15 @@ class FITSCutout(ImageCutout):
...
-    def __init__(self, input_files: List[Union[str, Path, S3Path]], coordinates: Union[SkyCoord, str], 
+
+    def __init__(self, input_files: List[Union[str, Path, S3Path]], coordinates: Union[SkyCoord, str],
                  cutout_size: Union[int, np.ndarray, Quantity, List[int], Tuple[int]] = 25,
                  fill_value: Union[int, float] = np.nan, limit_rounding_method: str = 'round',
-                 extension: Optional[Union[int, List[int], Literal['all']]] = None, 
-                 single_outfile: bool = True, verbose: bool = False):
-        # Superclass constructor 
+                 extension: Optional[Union[int, List[int], Literal['all']]] = None,
+                 single_outfile: bool = True, verbose: bool = False, fsspec_kwargs: Optional[dict] = None):
```
snbianco: Can you add a description for this parameter to the class docstring?

snbianco (commented):
Looks good! I only had a few comments. Thanks again for this PR.

astrocut/fits_cutout.py
```diff
@@ -59,15 +61,15 @@ class FITSCutout(ImageCutout):
...
```
snbianco: Now that I think about it, I would prefer that verbose be the last argument for both this class and the two convenience functions. It should also probably be a keyword-only arg 🫤, since now if we flip the order it will technically be a breaking change. I'll plan to make the change in another PR to make verbose keyword only.

snbianco (approved):
Looks good!

PR #179, opened 2026-04-03
Roman Spectral Cutouts

snbianco:
ASDF (Roman) spectral cutouts!

I recommend starting with the docs/tests to get an idea of how the code works. Main docs are under Astrocut Documentation -> Spectral Cutouts. File format descriptions are under Astrocut File Formats -> ASDF Cutout Files -> Spectral Cutouts.

The `RomanSpectralCutout` is the main class at work here. It accepts one or more spectral files, one or more source IDs, and an optional wavelength range to filter by. Astrocut opens each spectral file, searches for each of the source IDs, and creates cutouts based on the wavelength range. To get the ASDF cutout objects in memory, we use the `get_asdf_cutouts` method; to write to disk, we use `write_as_asdf`. The class takes an optional `lite` parameter which controls whether or not the cutouts contain a minimal set of data and metadata.

The `roman_spectral_cut` method is a convenience function that calls `RomanSpectralCutout` with multiprocessing. For cases where a very large number of source IDs are requested, this method dramatically improves performance over serial execution. Because of limitations in pickling, this method can only write ASDF cutout files to disk and can only create cutouts grouped by source ID AND input file.


astrocut/asdf_spectral_cutout.py
```diff
@@ -0,0 +1,433 @@
...
+        if not self._out_trees:
+            raise InvalidQueryError(
+                "No cutouts were created. Please verify that the source IDs are present in "
+                "the spectral files and that the wavelength range overlaps with the data."
+            )
+
+    def get_asdf_cutouts(
+        self,
+        *,
+        group_by: Literal["source_file", "file", "combined"] = "source_file",
```
snbianco: Any thoughts on what should be the default option here?

scfleming: I think I like "combined" as a default: we should be thinking about making this best on the cloud/platform.  If asking for spectra from a number of objects spread across a variety of source files, would a single asdf "combined" object in memory be the best user experience to work with these in a notebook or script on Nexus?  If so, that sounds like the right choice for a default, since we want a motto of "if a user doesn't change anything will they avoid saving a file to local area in the cloud?"

astrocut/asdf_spectral_subset.py
```diff
@@ -0,0 +1,433 @@
...
+        Generate the spectral cutout(s) from the input ASDF files based on the specified
+        source IDs and wavelength range.
+    """
+
+    def __init__(
+        self,
+        spectral_files: Union[str, Path, S3Path, List[Union[str, Path, S3Path]]],
+        source_ids: Union[str, int, List[Union[str, int]]],
+        wl_range: Union[tuple, list] = None,
+        lite: Optional[bool] = True,
```
snbianco, line 357: Should `lite=True` be the default here? I lean towards yes, since we generally want the defaults to be as efficient as possible. I have the same question for `roman_spectral_cut()`.

scfleming:
@snbianco this should probably be done in a separate PR, and apologies for not saying this earlier, but I think we'll want to go through and avoid calling these "cutouts" in the source code, input arguments, and documentation.  People will assume a "spectral cutout" means "you started with a two-dimensional spectral image and did the extraction yourself to make a 1-d spectrum."  That isn't what is happening here, it's really a "selection" or "subselection" or some similar word, where we are just indexing/slicing/pulling a specific array(s) from a larger pile of arrays.

So I think in a separate, mostly maintenance PR, we should go through anywhere we use "cutout" in a spectral context and replace it with the word we want to use in practice.

astrocut/asdf_spectral_subset.py
scfleming: Do the `lite=True` versions of the cutouts still work with our other tool suite, like Jdaviz?  If so, I agree.  If not, we'll want the default to be able to be passed over to a tool like `mast-aladin` and `Imviz` and have it work.

astrocut/roman_spectral_cutout.py
```diff
@@ -0,0 +1,138 @@
...
+    lite: bool = True,
+    output_dir: Union[str, Path] = ".",
+    batch_size: int = 128,
+    workers: Optional[int] = None,
+) -> List[str]:
+    """
+    Generate and write Roman spectral cutouts in parallel using multiprocessing.
+
+    This function divides the source IDs into batches and processes them in parallel across multiple worker processes.
+    Cutouts are grouped by source ID and file so that each output ASDF file contains cutouts for a single source ID
```
snbianco: Note that this function only writes cutouts per source ID per input file. Because of limitations with pickling, we can't parallelize the other two types of groupings without files being overridden or corrupted.

snbianco:
> @snbianco this should probably be done in a separate PR, and apologies for not saying this earlier, but I think we'll want to go through and avoid calling these "cutouts" in the source code, input arguments, and documentation. People will assume a "spectral cutout" means "you started with a two-dimensional spectral image and did the extraction yourself to make a 1-d spectrum." That isn't what is happening here, it's really a "selection" or "subselection" or some similar word, where we are just indexing/slicing/pulling a specific array(s) from a larger pile of arrays.
> 
> So I think in a separate, mostly maintenance PR, we should go through anywhere we use "cutout" in a spectral context and replace it with the word we want to use in practice.

This is good to know! I see how "cutout" might be misleading here. This means we should replace the word cutout everywhere? Because if that's the case, and we need to change class/function names, that should probably be done in this PR.

scfleming:
> > @snbianco this should probably be done in a separate PR, and apologies for not saying this earlier, but I think we'll want to go through and avoid calling these "cutouts" in the source code, input arguments, and documentation. People will assume a "spectral cutout" means "you started with a two-dimensional spectral image and did the extraction yourself to make a 1-d spectrum." That isn't what is happening here, it's really a "selection" or "subselection" or some similar word, where we are just indexing/slicing/pulling a specific array(s) from a larger pile of arrays.
> > So I think in a separate, mostly maintenance PR, we should go through anywhere we use "cutout" in a spectral context and replace it with the word we want to use in practice.
> 
> This is good to know! I see how "cutout" might be misleading here. This means we should replace the word cutout everywhere? Because if that's the case, and we need to change class/function names, that should probably be done in this PR.

I think that would be preferred, assuming these classes and functions are indeed restricted to spectral support from SSC.  But then that also means we needed to decide the replacement word to use instead...

Not to be too cute, but, are these "snips" instead of "cutouts", keeping to the general theme?  I think we definitely want to use the words "extract" or "extraction".

"Indexing" or "slicing" might be (roughly) correct technically, but might be confusing for a general audience.  "Subsetting" already has a defintion/concept with the FITS work to select a small section of a FITS data structure, so maye that is reasonable here too?

snbianco:
I would think that "extraction" might also cause confusion, since users might think we mean to extract a spectrum from an image. `SpectralSnip` would be fun, but maybe not as descriptive. Maybe something like `SpectralSelection` or `SpectralSubset`.

scfleming:
> I would think that "extraction" might also cause confusion, since users might think we mean to extract a spectrum from an image. `SpectralSnip` would be fun, but maybe not as descriptive. Maybe something like `SpectralSelection` or `SpectralSubset`.

Yes I am leaning towards one of those too, am curious what others think.

astrocut/asdf_spectral_cutout.py
havok2063: I think that makes sense for the end user, especially for the RRN.  For the Roman MM Extracted Search, it will depend a bit on the desire.  For users who want to select and download rows, likely "combined" into a single download file is best.  But for loading rows into an embedded Specviz, we'd want "source_file" + "in-memory" as the output.

astrocut/asdf_spectral_subset.py
havok2063: Currently, neither the `lite` version or the "full" cutout work in Specviz.    The `lite` version is minimally missing  `unit_flux` and `unit_wl` from the meta dictionary and the `flux_error` array from the data.  I would probably just include the full `meta` dict even in the `lite` version for provenance info.  

The cutouts don't load into Specviz because the app currently expects the original SSC format of the spectral data nested under the source_id key within `data`.   For single source spectra, I like having the spectral data right under `data`.  We'll need to push updates to Jdaviz and specutils to explicitly understand the cutout format.

havok2063:
I like `SpectralSubset`.  It's also in line with the language that Jdaviz uses when slicing by wavelength.  I think we will want to make spectral cutouts more generally available.  It would be useful for MaNGA or JWST IFU cubes.

astrocut/asdf_spectral_cutout.py
havok2063: For the "source_file" group_by, it's a bit weird to have a tuple as a dict key.  It makes accessing the values a bit more cumbersome. 
```
{('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '402849'): <asdf._asdf.AsdfFile at 0x13f7ad0d0>,
 ('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '404979'): <asdf._asdf.AsdfFile at 0x1663bbce0>}
```
I would opt to either just have the "source_id" as key, or use the same file stem as when you write out the file, e.g. `wfi_spec_combined_1d_r0000201001001001001_0002_WFI01_cutout_402849_lite`

havok2063:
I checked that the SSC `individual_1d` files work with the current code.  It works at least with the default values.  The `decontanimated_2d` files do NOT work right now, because they don't have a `wl` array.  According to the datamodel, they should have one.  I think these are supposed to be cutouts resampled onto a common wavelength grid.   I think we can defer support for this to later until we get newer files or confirmation from SSC.

havok2063:
Another question is if we want to support source cutouts from both PRISM and GRISM files at the same time?  Or restrict it to just one instrument at a time?  It does seem to work, but it could lead to strange behaviour, that we might need to account for.   Here is an output of a test where I gave it a prism and grism file and asked it to extract 2 sources from the prism and 1 from the grism.  The 2 prism sources were also in the grism file, which I wasn't expecting. 

```
gsc = RomanSpectralCutout(spectral_files=gspec_files, source_ids=[402849, 404979, 184839], lite=True)
gsc.get_asdf_cutouts()
WARNING: DataWarning: Source ID 184839 not found in file wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf. Skipping this source for this file. [astrocut.asdf_spectral_cutout]
Out[129]:
{('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '402849'): <asdf._asdf.AsdfFile at 0x167c4b830>,
 ('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '404979'): <asdf._asdf.AsdfFile at 0x167c45cd0>,
 ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
  '184839'): <asdf._asdf.AsdfFile at 0x167c52a80>,
 ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
  '402849'): <asdf._asdf.AsdfFile at 0x167c05550>,
 ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
  '404979'): <asdf._asdf.AsdfFile at 0x167c2ab40>}
```
Currently, the SSC grism and prism filenames have the same name.    My hunch is to restrict it.

astrocut/roman_spectral_cutout.py
```diff
@@ -0,0 +1,138 @@
...
+
+    cutout = RomanSpectralCutout(
+        spectral_files=filepath, source_ids=source_id_batch, wl_range=wl_range, lite=lite, verbose=False
+    )
+
+    return cutout.write_as_asdf(output_dir)
+
+
+def roman_spectral_cut(
+    spectral_files: Union[str, Path, S3Path, List[Union[str, Path, S3Path]]],
```
havok2063: I ran this with 2 input files, 1 small file and 1 file with 100,000 rows.  I asked it to extract 2 sources from the small file and 2 from the large one.  It took 16 seconds to run.  This was the same time as when using `RomanSpectralCutout.get_asdf_cutouts` directly in a python session.   This is probably in opening the large file once.  

I think we'll want to test this with several large files (5-10) of 100,000 rows and see how this performs.  If it takes 16 seconds to open each file, this time could grow significantly.

astrocut/asdf_spectral_cutout.py
```diff
@@ -0,0 +1,433 @@
...
+                source_ids = [sid for sid in sources_to_include if sid in self._out_trees[file]]
+                if self._lite:
+                    combined_tree = {
+                        self._mission_keyword: {
+                            "data": {
+                                sid: self._out_trees[file][sid][self._mission_keyword]["data"]
+                                for sid in sources_to_include
+                                if sid in self._out_trees[file]
+                            },
+                            "meta": {"source_ids": source_ids},
```
havok2063: I think we should include more than just the source ids here, at the very least for provenance.   This could be confusing when combining all sources into a single file.  Maybe we want to keep track of a simple mapping between source id and original cutout file?

astrocut/asdf_spectral_cutout.py
```diff
@@ -0,0 +1,433 @@
...
+                        f"Spectral cutout created for source ID {sid} from file {file} "
+                        f"with wavelength range {self._wl_range}.",
+                        software={
+                            "name": "astrocut",
+                            "author": "Space Telescope Science Institute",
+                            "version": __version__,
+                            "homepage": "https://astrocut.readthedocs.io/en/latest/",
+                        },
+                    )
+                    asdf_cutouts[(file, sid)] = af
```
havok2063: See other comment on the tuple as a dict key.

astrocut/asdf_spectral_cutout.py
```diff
@@ -0,0 +1,433 @@
...
+                    key_by_file = {
+                        str(file): self._base_trees[file][key]
+                        for file in files_to_include
+                        if key in self._base_trees[file]
+                    }
+                    base_tree[f"{key}_combined"] = key_by_file
+
+                meta_by_file = {}
+                for file in files_to_include:
+                    meta_by_file[str(file)] = {
```
havok2063: This also applies elsewhere, but what happens if the input files have the same name?

havok2063: Well nevermind.  It saves the full input path, including relative paths.

astrocut/asdf_spectral_cutout.py
snbianco: Yeah, I agree that its inconvenient for the user to have to index by a tuple. I'm really not sure of a way to get around it though, unless we make it a nested dictionary. I'm not sure that indexing by the file stem would be any easier for a user, since they would have to know the pattern. At least with a tuple, you could use variables already set, like this:
```
source_ids = [1, 2, 3]
files = ["file1", "file2"]
subset = SpectralSubset(files, source_ids)
afs_dict = subset.get_asdf_cutouts()
af = afs_dict[(source_ids[0], files[0])]
```

astrocut/asdf_spectral_subset.py
snbianco: I'll update to have `lite` versions include full metadata and the `flux_error` array in data.

snbianco:
> Another question is if we want to support source cutouts from both PRISM and GRISM files at the same time? Or restrict it to just one instrument at a time? It does seem to work, but it could lead to strange behaviour, that we might need to account for. Here is an output of a test where I gave it a prism and grism file and asked it to extract 2 sources from the prism and 1 from the grism. The 2 prism sources were also in the grism file, which I wasn't expecting.
> 
> ```
> gsc = RomanSpectralCutout(spectral_files=gspec_files, source_ids=[402849, 404979, 184839], lite=True)
> gsc.get_asdf_cutouts()
> WARNING: DataWarning: Source ID 184839 not found in file wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf. Skipping this source for this file. [astrocut.asdf_spectral_cutout]
> Out[129]:
> {('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
>   '402849'): <asdf._asdf.AsdfFile at 0x167c4b830>,
>  ('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
>   '404979'): <asdf._asdf.AsdfFile at 0x167c45cd0>,
>  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
>   '184839'): <asdf._asdf.AsdfFile at 0x167c52a80>,
>  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
>   '402849'): <asdf._asdf.AsdfFile at 0x167c05550>,
>  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
>   '404979'): <asdf._asdf.AsdfFile at 0x167c2ab40>}
> ```
> 
> Currently, the SSC grism and prism filenames have the same name. My hunch is to restrict it.

I don't see the harm in allowing users to cut out from both at the same time provided that they point to two different files. All of those IDs are in the grism file, so I think it makes sense that there would be a cutout for each of them.

astrocut/roman_spectral_cutout.py
snbianco: Yeah, I would guess that the vast majority of that time is in opening the file. I just ran a test with three different xlarge files, and it took around 13.2 seconds on my machine. You don't really start to see any performance improvements over the class until you're asking for a very large amount of cutouts to be written to disk. I mention in the documentation that this function isn't recommended for a small number of subsets.

astrocut/asdf_spectral_cutout.py
snbianco: I'll change the `lite` cutouts to include all of the original metadata. For the combined ASDF objects, meta is keyed by input file, and each input file has its own set of source IDs.

scfleming:
> > Another question is if we want to support source cutouts from both PRISM and GRISM files at the same time? Or restrict it to just one instrument at a time? It does seem to work, but it could lead to strange behaviour, that we might need to account for. Here is an output of a test where I gave it a prism and grism file and asked it to extract 2 sources from the prism and 1 from the grism. The 2 prism sources were also in the grism file, which I wasn't expecting.
> > ```
> > gsc = RomanSpectralCutout(spectral_files=gspec_files, source_ids=[402849, 404979, 184839], lite=True)
> > gsc.get_asdf_cutouts()
> > WARNING: DataWarning: Source ID 184839 not found in file wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf. Skipping this source for this file. [astrocut.asdf_spectral_cutout]
> > Out[129]:
> > {('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
> >   '402849'): <asdf._asdf.AsdfFile at 0x167c4b830>,
> >  ('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
> >   '404979'): <asdf._asdf.AsdfFile at 0x167c45cd0>,
> >  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
> >   '184839'): <asdf._asdf.AsdfFile at 0x167c52a80>,
> >  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
> >   '402849'): <asdf._asdf.AsdfFile at 0x167c05550>,
> >  ('../g2dp_ver3.3_grism/wfi_spec_combined_1d_r0000101001001001001_0001_WFI01.asdf',
> >   '404979'): <asdf._asdf.AsdfFile at 0x167c2ab40>}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Currently, the SSC grism and prism filenames have the same name. My hunch is to restrict it.
> 
> I don't see the harm in allowing users to cut out from both at the same time provided that they point to two different files. All of those IDs are in the grism file, so I think it makes sense that there would be a cutout for each of them.

If they have the same Object IDs and we aren't capturing other metadata though, we may need to include some other metadata fields in the output to differentiate WHICH spectrum is which?

havok2063:
> If they have the same Object IDs and we aren't capturing other metadata though, we may need to include some other metadata fields in the output to differentiate WHICH spectrum is which?

Yeah that's what I was thinking.  Now that the full meta is included in the output of the `lite` this could have enough info for users to differentiate between the two, besides the filename.  The metadata includes `'optical_element': 'PRISM'`, although I don't know if this is standard Roman metadata or an SSC temporary field.

astrocut/roman_spectral_cutout.py
havok2063: Ok, thanks for clarifying.  I must have missed that in the docs.

scfleming:
> The metadata includes `'optical_element': 'PRISM'`, although I don't know if this is standard Roman metadata or an SSC temporary field.

`optical_element` should be both a standard SOC and SSC metadata keyword, and exposed via AUI, etc.  Values are GRISM or PRISM for spectra.

astrocut/asdf_spectral_cutout.py
havok2063: I'm not sure I follow?  Why does it need to be a nested dict?  A string key could still be made as a unique index.  I think a lot of users will be inspecting the output and doing copy/paste of a key from one cell to another.  IMO a string is a bit easier to do that than a tuple, and can also be programmatically created.    A tuple almost requires a user to programmatically build the key all the time.   And they'll have to know the right source + file combinations. 

Something like this feels a bit easier to digest...
```
{
 '402849_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01' : <asdf._asdf.AsdfFile>,
 '404979_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile>
}
``` 
than 
```
{('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '402849'): <asdf._asdf.AsdfFile at 0x13f7ad0d0>,
 ('wfi_spec_combined_1d_r0000201001001001001_0002_WFI01.asdf',
  '404979'): <asdf._asdf.AsdfFile at 0x1663bbce0>}
``` 
We might want to create and store the keys used for selection as a variable on `subset`, or provide a method to generate the unique keys.  This would be available so the user doesn't have to programmatically make the key or know which source+file combinations are valid.   This would be useful in any case but especially if we keep the tuple as key.

snbianco:
I just pushed some new changes!

- `RomanSpectralSubset`
  - Error checking for an invalid `group_by` value.
  - Objects grouped by `source_file` are now keyed by a string rather than a tuple. Added a `get_source_file_keys` function that returns a dictionary of string keys mapped to their corresponding tuple.
- `roman_spectral_subset`
  - Added `group_by` as an option. "File" creates a set of jobs divided by input file. "Combined" does no parallelization at all and calls the worker directly. 
  - Changed the default `batch_size` and `workers` based on my benchmark testing.
  - Better input validation for spectral files and source IDs.
  - Optimized grouping by `source_file` to better balance the "buckets" of source ID based on the number of workers available.

snbianco:
Something that just occurred to me is that if we key `source_file` by a string, we might run into duplicates for files that have the same stem (like prism vs. grism) 😬

havok2063:
> * Objects grouped by `source_file` are now keyed by a string rather than a tuple. Added a `get_source_file_keys` function that returns a dictionary of string keys mapped to their corresponding tuple.

I like the new `get_source_file_keys()` and `make_source_file_key` methods.  I think they will be helpful.

> Something that just occurred to me is that if we key `source_file` by a string, we might run into duplicates for files that have the same stem (like prism vs. grism) 😬

Yeah, that's true.  I don't think they can have the same file stem unless they are served from different directories, in which case you'd get the last one loaded.  We could extract the `optical_element` field from the meta and affix it to the source key.   Something like...
```
{'402849_prism_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile at 0x107095340>,
 '404979_prism_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile at 0x14f972240>,
 '184839_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x1072eb440>,
 '402849_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x14f9b9bb0>,
 '404979_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x14f87faa0>}
```
What do you think of that?  Also, if you have a preference for having the stem first and suffxing it with the source id, etc,  feel free to change it. I don't have a strong preference there.

snbianco:
> > * Objects grouped by `source_file` are now keyed by a string rather than a tuple. Added a `get_source_file_keys` function that returns a dictionary of string keys mapped to their corresponding tuple.
> 
> I like the new `get_source_file_keys()` and `make_source_file_key` methods. I think they will be helpful.
> 
> > Something that just occurred to me is that if we key `source_file` by a string, we might run into duplicates for files that have the same stem (like prism vs. grism) 😬
> 
> Yeah, that's true. I don't think they can have the same file stem unless they are served from different directories, in which case you'd get the last one loaded. We could extract the `optical_element` field from the meta and affix it to the source key. Something like...
> 
> ```
> {'402849_prism_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile at 0x107095340>,
>  '404979_prism_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile at 0x14f972240>,
>  '184839_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x1072eb440>,
>  '402849_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x14f9b9bb0>,
>  '404979_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x14f87faa0>}
> ```
> 
> What do you think of that? Also, if you have a preference for having the stem first and suffxing it with the source id, etc, feel free to change it. I don't have a strong preference there.

My only concern with this is the case in which a user has two different input files with the same optical element in two different directories that are named the same thing. For example, a user having two prism files each called something like `spectrum.asdf`. This would be a corner case, so we probably wouldn't run into it too often. One solution might be to add in the optical element to the key string, and if a duplicate is still detected, append a hash or even use the full relative filename to better differentiate.

havok2063:
> My only concern with this is the case in which a user has two different input files with the same optical element in two different directories that are named the same thing. For example, a user having two prism files each called something like `spectrum.asdf`. This would be a corner case, so we probably wouldn't run into it too often. One solution might be to add in the optical element to the key string, and if a duplicate is still detected, append a hash or even use the full relative filename to better differentiate.

Yeah I agree.   I like appending a simple hash to force key uniqueness for cases where it's still ambiguous.  If users wanted to know which one was which, they could run `get_source_file_keys()` to see the mapping between the source key and the input tuple, which would contain the full relative path.

snbianco:
Do we know exactly what the filenames will look like for Roman data products yet? If they start with `grism` or `prism`, it may not even be worth adding the optical element to all keys and we could just add the disambiguating hash to those that need it. I'm specifically looking at this example output:
```
{'426524_prism_large_wfi_spec_combined_1d_r0000201001001001001_0002_WFI01': <asdf._asdf.AsdfFile at 0x32889e590>,
 '82038_grism_grism_wfi_spec_combined_1d_r0000101001001001001_0001_WFI01': <asdf._asdf.AsdfFile at 0x32bc6fad0>}
```

scfleming:
> Do we know exactly what the filenames will look like for Roman data products yet? If they start with `grism` or `prism`, it may not even be worth adding the optical element to all keys and we could just add the disambiguating hash to those that need it.

Yes we do, for the single-visit 1D asdf files:
`r<ppppp>_<product release identifier>_<cc><aaa><sss><ooo><vvv>_<eeee>_<ssc_skycell>_<optical element>_individual-1d.asdf`

For the multi-visit combined 1D asdf files:
`r<ppppp>_<product release identifier>_<ssc_skycell>_<optical element>_<g2dp_run_id>_combined-1d.asdf`

In both cases, the source files will have "prism" or "grism" in their file names in the <optical_element> field.  If you need the full breakdown of what these fields mean, I can get you the info at least as it stands for Build 22.

havok2063:
> Yes we do, for the single-visit 1D asdf files: `r<ppppp>_<product release identifier>_<cc><aaa><sss><ooo><vvv>_<eeee>_<ssc_skycell>_<optical element>_individual-1d.asdf`
> 
> For the multi-visit combined 1D asdf files: `r<ppppp>_<product release identifier>_<ssc_skycell>_<optical element>_<g2dp_run_id>_combined-1d.asdf`

Oh nice.  Then we shouldn't need to explicitly add anything except a possible hash at the end.    Are there example files available that use that naming schema?  Or should we continue to operate with the v3.3 files we have?

scfleming:
> Are there example files available that use that naming schema?

We have one sample of each, although I can't speak to the contents of the spectra themselves, but they should have the correct file names.  I'll send you the location in our work Slack.

havok2063:
I've checked that both the individual and the combined have the same file structure as the original 3.3 files we were using for testing.

snbianco:
Upon further experimentation, I've realized that I can parallelize the file opening in `RomanSpectralSubset.subset` and the file writing in `RomanSpectralSubset.write_as_asdf`. Having the parallelization directly within the class has the added benefit of parallelizing large file openings, even when not writing subsets to disk. These two functions together would essentially do the same thing as `roman_spectral_subset`,  but there is some convenience in having a function that puts these things together. It's also possible that having two different entry points would just be confusing for users. Thoughts? 

I lean more towards keeping things lightweight and removing `roman_spectral_subset()` in favor of parallelization parameters inside the main class. The default would be to process sequentially since overhead generally outweighs any performance benefit in a lot of cases.

havok2063:
I'd be ok consolidating things down, and removing `roman_spectral_subset`, if it makes sense to do so.   Would the user code then be something like 
```
sc = RomanSpectralSubset(...)
sc.write_to_asdf()
```
with a keyword flag to enable parallelization?  That's simple enough.  

Is there a downside to making it the default option?   If so, I think we'd want some documentation on when, and when not, to parallelize.  And if we have ballpark file numbers and file sizes to give as a guide, that might be helpful for users.  For use in MAST Roman MM, would we default to parallelize?

snbianco:
Just pushed some changes. I did a lot of experimenting, and weirdly enough, parallelizing `write_as_asdf` and `get_asdf_cutouts` has virtually no performance improvement, even when writing thousands of different subsets. I think the overhead per file dominates.

Because of this, I chose to leave `write_as_asdf` and `get_asdf_cutouts` as is and only parallelize `_subset_file` to be able to open different input files at the same time. That has a very clear performance improvement, especially for very large files. I can try to do some analysis on file numbers/sizes where parallelizing has an advantage.

havok2063:
> Just pushed some changes. I did a lot of experimenting, and weirdly enough, parallelizing `write_as_asdf` and `get_asdf_cutouts` has virtually no performance improvement, even when writing thousands of different subsets. I think the overhead per file dominates.

That's interesting.  Was this with all three `group_by` options?   Do you see the same when writing out to individual source files?

snbianco:
Yeah, writing with `group_by="source_file"` was the majority of the testing I did. I think in that case, the files/objects are so small that parallelization doesn't really do much. `combined` is inapplicable since it only writes a single file. I didn't see any speed up for `file` mode either, but maybe there may be more of a difference for a huge amount of input files. The most I checked was 10.

snbianco:
> Yeah, writing with `group_by="source_file"` was the majority of the testing I did. I think in that case, the files/objects are so small that parallelization doesn't really do much. `combined` is inapplicable since it only writes a single file. I didn't see any speed up for `file` mode either, but maybe there may be more of a difference for a huge amount of input files. The most I checked was 10.

I have an amendment to this! I was also curious to know why parallel execution wouldn't help at all, so I tried again with some very bulky file writes. There seems to be no improvement up until around 10,000 file writes, and then you start seeing some advantage in using a process pool with batching. I can add this as an option to `write_as_asdf` when `group_by="source_file"`, and do some benchmarking on that as well to see when using processes actually becomes most worthwhile.

docs/astrocut/index.rst
```diff
@@ -357,16 +357,25 @@ subsets in different ways. They accept the following parameters:
...
-`~astrocut.RomanSpectralSubset` also accepts a ``max_workers`` parameter that controls file-level multiprocessing during
+`~astrocut.RomanSpectralSubset` accepts a ``max_workers`` parameter that controls file-level multiprocessing during
 subset generation:
 
-- ``max_workers=1`` (default): process files sequentially.
-- ``max_workers=None``: automatically choose a worker count based on available CPUs and number of input files.
-- ``max_workers>1``: process multiple input files in parallel with that worker count.
+- ``max_workers=1`` (default): Process files sequentially.
+- ``max_workers=None``: Automatically choose a worker count based on available CPUs and number of input files.
+- ``max_workers>1``: Process multiple input files in parallel with that worker count.
```
havok2063: I wonder if `max_workers=None` should be the default.   I'm thinking that most files we'll get will have >10,000 sources in them.   We won't know for sure until we get real data, but we start seeing speed-ups as low as 1000 sources.  And there's not much performance hit until really low source numbers.  The number of files a user inputs I think will depend heavily on how and where they are searching for data.   

For use on the Nexus, and probably in MissionMast as well, I think we'll want to minimize memory usage,  file storage space, and the subset time.

snbianco: This is good with me if we can safely assume that most input files will be large. The most recent commit implements this as well as multiprocessing in the `write_as_asdf` function. I defaulted to 1 worker there, since we generally want to discourage users from writing a lot of files, and the default `group_by` value will be combined. 

I also implemented a parameter that bypasses the ASDF validation check when writing to files. This is maybe not ideal, but that check has a major impact on write time of files. Let me know if this is a horrible idea that I should scrap immediately. If not horrible, then what value should this default to? Do we stick to the better practice `validate_output=True` or the option that minimizes compute `validate_output=False`.

havok2063: What does the ASDF validation check do?  Is this something internal to the `asdf` package?  Does it check the data against the schema definition or something like that?   Is this something we or the user can easily upset?  

On your relative throughput plot, should I read that as when validate is False, parallelized workers vs single workers are about the same, and when validate is True, parallelized workers give up to 3x speedup over sequential? 

It's not a crazy idea to have it as an option.  On the one hand, we should know with some certainty the structure of the input files, and we'll control the output structure.  Also, in a future iteration of the tool, I think we'll handle the source_id -> file lookup internally, so the user doesn't have to pass in files, which further ensures good inputs.     On the other hand, users could still muck with the input file structure, but then the code wouldn't work, and this seems unlikely.  

I think it's best to probably leave `validate=True` as the default.   Another option is when validate is False, to still validate at least one file before writing out all of them.  And fail if that validation fails.   We can document that if a user is trying to write out thousands of file, they can turn off asdf validation manually to improve write times.

havok2063: From the other validation plots, it also looks like there's very little validation overhead when writing with parallelization.  Is that right?  If so, another reason maybe to make max_workers=None the default.

snbianco: The ASDF validation check is internal to the `asdf` package. It validates the provided ASDF tree against the appropriate schema based on the tree's tag. I agree with all you said about how likely it is for us or a user to "break" a file. We don't change any data types, and for full cutouts, the majority of the data/metadata is preserved from input to subset. The one thing I might worry about is schema changing for future tags, but we should be able to catch that in the regression tests.

Yes, your reading of the plot is correct. There's a huge benefit to using parallelization when validate is True. I do like the idea of just validating a single file as a sanity check.

The worry I have with making `max_workers=None` the default is the case where the user is asking for more than one but less than 1,000 files. Although looking back at the benchmark data, the parallel write for 100 files still runs in about 4 seconds, and the user has the option of manually setting `max_workers`. It's probably more important to default to the 3x speedup for larger file writes. The default of the "combined" grouping setting will be serial in all cases since it's only writing one file, so I think I agree that `max_workers=None` is the way to go.

havok2063: That all makes sense to me.   Plus I think if the code finds a mismatch between the asdf schema version and the input files, such that it can't read/write the data, it should fail gracefully and inform the user.  

You make a good point about `max_workers` and the number of files.   The `combined_1d` file looks organized by sky cell.  I think if a user is doing a cone search then they'll likely only intersect maybe a few 10s of files at most.  If they are doing a Catalog Search or MissionMast search with parameters, then they could potentially get a set of sources across many files.   All of those search options will be available with astroquery on the Nexus.   I guess in one case sequential would be better and the other parallelized?  

If we left the defaults at `group_by=combined` and `max_workers=None`, could we configure max_workers to smartly switch to 1 or None based on the identified number of sources and files?   I'm just thinking out loud a bit here.  Overall I think this PR is looking real good.

snbianco: I like that idea! Latest commit also takes the number of write jobs into account when determining worker count if `max_workers` is set to None.

havok2063 (approved):
I think this is in good shape, and feel any new updates can happen as refinements in a new PR.  I'll approve, but if you have more you want to do, then feel free to keep working on it.

PR #181, opened 2026-04-29
Switch to ruff for linting

snbianco:
Switching to use `ruff` in our codestyle checks. Also updating all the current files with ruff formatting.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,7 +51,27 @@ For active development install in develop mode
 .. code-block:: bash
 
     $ pip install -e .
-    
+
+Linting
+=======
+Astrocut uses `ruff <https://github.com/charliermarsh/ruff>`_ for linting.
+
+To run ruff, install it with ``pip install ruff`` and then run the following commands from the root of the repository:
+
+.. code-block:: bash
+
+    $ ruff check .
+    $ ruff format --check .
+
+To automatically fix linting issues, run:
+
+.. code-block:: bash
+
+    $ ruff check . --fix
+    $ ruff format .
+
+Be careful with the above commands, as they will make changes to your code. You should review the changes before committing them.
+
 Testing
 =======
 Testing is now run with `tox <https://tox.readthedocs.io>`_ (``pip install tox``).
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ For active development install in develop mode
 
 Linting
 =======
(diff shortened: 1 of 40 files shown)


tox.ini
```diff
@@ -49,24 +35,17 @@ deps =
...
-    test
-    alldeps: docs
 
 commands =
     pip freeze
-    !cov: pytest --pyargs astrocut {toxinidir}/docs {posargs}
-    cov: pytest --pyargs astrocut {toxinidir}/docs --cov astrocut --cov-config={toxinidir}/setup.cfg {posargs}
-    cov: coverage xml -o {toxinidir}/coverage.xml
+
+    cov: pytest --cov=astrocut --cov-config={toxinidir}/setup.cfg --cov-report=term-missing --cov-report=xml:{toxinidir}/coverage.xml {toxinidir}/astrocut {toxinidir}/docs {posargs}
```
falkben: why was `--pyargs astrocut` dropped?

tox.ini
```diff
@@ -1,8 +1,8 @@
 [tox]
 envlist =
-    py{38,39,310,311,312,313}-test{,-alldeps,-devdeps}{,-cov}
-    py{38,39,310,311,312,313}-test-numpy{120,123}
-    py{38,39,310,311,312,313}-test-astropy{52}
+    py{39,310,311,312,313,314}-test{,-alldeps,-devdeps}{,-cov}
```
falkben, line 3: maybe worth calling out in a changelog that we're no longer supporting py3.8 (and we're adding support for 3.14), though i fully support this change.

README.rst
```diff
@@ -51,7 +51,27 @@ For active development install in develop mode
 .. code-block:: bash
 
     $ pip install -e .
-    
+
+Linting
+=======
+Astrocut uses `ruff <https://github.com/charliermarsh/ruff>`_ for linting.
```
falkben:
```suggestion
Astrocut uses `ruff <https://github.com/astral-sh/ruff>`_ for linting.
```

falkben (commented):
This makes me so happy!

tox.ini
snbianco: That's my bad, it shouldn't have been. I was experimenting with how to pass extra arguments to pytest and forgot to add it back! I'll replace in the next commit.

falkben (approved):

PR #182, opened 2026-04-29
Remove TICA completely

snbianco:
Fully removing TICA cubes and cutouts from Astrocut after we deprecated in 1.1.0.

I ran into some trouble with rebasing and accidentally closed the other PR.


astrocut/tests/test_cube_cut.py
```diff
@@ -134,9 +122,10 @@ def test_cube_cutout(cube_file, ffi_files, ffi_type, use_factory, tmp_path):
     num_im = 100
 
     # Read one of the input images to get the WCS
-    n = 1 if ffi_type == "SPOC" else 0
+    n = 1
     img_header = fits.getheader(ffi_files[0], n)
     with warnings.catch_warnings():
+        warnings.simplefilter("ignore", FITSFixedWarning)
```
falkben: duplicate line

```suggestion
```

astrocut/tests/test_cube_cut.py
```diff
@@ -390,23 +348,21 @@ def test_target_pixel_file(cube_file, ffi_type, tmp_path):
...
-    assert len(tpf_table.columns) == ncols
+    assert len(tpf_table.columns) == 12
     assert "TIME" in tpf_table.columns.names
     assert "FLUX" in tpf_table.columns.names
     assert "FLUX_ERR" in tpf_table.columns.names
     assert "FFI_FILE" in tpf_table.columns.names
 
     # Check img cutout shape and data type
     cutout_img = tpf_table[0]["FLUX"]
+    cutout_img = tpf_table[0]["FLUX"]
```
falkben: duplicate


```suggestion
```

astrocut/tests/test_tess_footprint_cutout.py
```diff
@@ -38,7 +38,7 @@ def coordinates():
 
 
 @pytest.fixture
-def all_ffis(scope="module"):
+def all_ffis():
```
falkben, line 41: why was the scope changed here?

falkben (changes requested):

astrocut/tests/test_tess_footprint_cutout.py
snbianco: I don't think the scope was placed right here - it's supposed to be an argument of the fixture statement. Since it wasn't actually setting the scope in the first place, I opted to just remove it.

astrocut/tests/test_cube_cut.py
snbianco: Ah good catch! Probably a leftover of my rebasing struggle.

astrocut/tests/test_tess_footprint_cutout.py
falkben: oh, good point, i didn't catch that. I still think the fixture probably _could_ be set to module scope if you wanted. Would make the tests run slightly faster, I imagine.

falkben (approved):

PR #184, opened 2026-05-12
ASDF Cutout bugfixes

snbianco:
Previously, ASDF cutouts including data arrays with three or more dimensions were not performed properly when inputting an angular (non-pixel) cutout size. Fix this bug and add a test case.


setup.cfg
```diff
@@ -39,7 +39,7 @@ test =
 docs =
     sphinx != 4.1.0
     docutils == 0.16
-    sphinx-astropy
+    sphinx-astropy==1.10
```
snbianco: The new version of `sphinx-astropy` (1.11) requires [astropy-sphinx-theme](https://github.com/astropy/astropy-sphinx-theme) version 3, which doesn't support Python 3.9.

havok2063: Do we still need to support python 3.9?

astrocut/asdf_cutout.py
```diff
@@ -369,19 +369,21 @@ def _get_cutout_data(self, tree: dict, wcs: WCS, pixel_coords: Tuple[int, int])
...
-                new_shape = obj.shape[:-2] + (self._cutout_size[1], self._cutout_size[0])
-                cutout_cube = np.full(new_shape, self._fill_value, dtype=obj.dtype)
+                cutout_cube = None
 
                 for idx in np.ndindex(obj.shape[:-2]):
-                    cutout = self._make_cutout(obj[idx], pixel_coords, None)
+                    cutout = self._make_cutout(obj[idx], pixel_coords, wcs)
+                    if cutout_cube is None:
+                        new_shape = obj.shape[:-2] + cutout.data.shape
+                        cutout_cube = np.full(new_shape, self._fill_value, dtype=cutout.data.dtype)
```
havok2063, line 399: can you clarify what this change is doing?

setup.cfg
scfleming: I suggest we drop it at this point.

astrocut/asdf_cutout.py
snbianco: Previously, it was determining the shape of the cube before making any cutouts. This incorrectly assumed that cutout size would always be in pixels. Now, we wait to make the cube until one plane's cutout is created, and use the size of the cutout (in pixels) to determine the shape of the cube.

setup.cfg
snbianco: Created a new PR to drop 3.9 at #185 . I'll wait until that goes in before merging this one.

snbianco:
Latest commit adds another fix for a bug that I recently came across where integer arrays that require a fill value error out of `Cutout2D` if the fill value is `np.nan`. I made it so that, for integer arrays, the fill value must be an integer. If conversion fails, it defaults to 0.

astrocut/tests/test_asdf_cutout.py
```diff
@@ -103,7 +103,7 @@ def test_images(tmp_path, fakedata):
         "roman": {
             "meta": meta,
             "data": data,
-            "dq": data,
+            "dq": data.astype(int),  # DQ is typically an integer array
```
havok2063, line 106: Do we need to worry about any cases where the DQ array has NaN or None values?

snbianco: That array is the bitwise sum of data quality flags, so I don't believe so. Also, integer arrays in `numpy` shouldn't be able to support NaN or None values.

snbianco:
Last commit was some changes to `tox.ini` that I somehow forgot to include on #185.

astrocut/asdf_cutout.py
havok2063: Can we add a comment here just noting that, for posterity?

havok2063 (dismissed):
I just added a comment to add a comment.  Otherwise, I think this looks good, so I'll approve now.

PR #185, opened 2026-05-14
Drop support for Python 3.9, some docs fixes

snbianco:
Dropping support for Python 3.9, fixing up some docs that were failing the link check, and also fixing up some issues with the new version of `sphinx_astropy`. I also took this opportunity to organize our CI tests a bit more.


havok2063 (approved):
Seems reasonable to me.

PR #186, opened 2026-05-19
Improvements for ASDF image cutouts

snbianco:
Improvements to ASDF cutouts, including:

- Default to `lite=True`.
- Better handling to check if `asdf_in_fits` is supported.
- Only import `requests` and `s3fs` if an S3 product is requested.
- Remove deep copies and only load in necessary arrays from the input file.
- Only call `Cutout2D` once for the data array, and then apply the cutout slices to other arrays (when `lite=False`).
- Only slice GWCS objects once and then cache.
- Cache a primary header template for FITS outputs and a fill value cache based on input data types.
- Option to not validate outputs when writing cutouts to ASDF files.


snbianco:
@braingram 

If you have time to look over this module, I would love to get your thoughts! Specifically, are we making cutouts from ASDF images in the most optimal way?

astrocut/asdf_cutout.py
```diff
@@ -1,30 +1,43 @@
...
+from packaging.version import Version
 from s3path import S3Path
 
 from . import __version__, log
 from .exceptions import DataWarning, InvalidInputError, InvalidQueryError, ModuleWarning
 from .image_cutout import ImageCutout
 
 
+@contextmanager
+def _disabled_asdf_validation():
```
braingram: Validation is a key feature of ASDF and disabling it seems problematic. It risks producing invalid, unreadable files. Testing with a roman level 2 file I only see a 1 millisecond difference when validation is disabled vs enabled. This seems like an insignificant performance difference for a considerable amount of risk.

scfleming:
Do you have any measurements of overall performance improvements with these changes?

astrocut/asdf_cutout.py
havok2063: Sam can correct me here, but I think when investigating the Roman ASDF spectral products, we found enabling validation had a significant impact on overall performance when writing out many files, ~3x slower with validation enabled when writing >10,000 files with a single processor.  But yeah, I think the difference in write was negligible for a small number of files.   But I don't think we've yet done a full performance comparison for image cutouts.

havok2063:
I wonder if it's worth, ahead of any future code changes, to run some performance testing on a bunch of use cases with the current code, to see how we do, then use that to pinpoint where we can make improvements?

scfleming:
> I wonder if it's worth, ahead of any future code changes, to run some performance testing on a bunch of use cases with the current code, to see how we do, then use that to pinpoint where we can make improvements?

Yes this profiling of the different scenarios is exactly the request of @snbianco to map out the test matrix, which she's working on this Sprint (please help out as you can.)  There's a number of variables here, including:
- are you storing the output in memory or writing to files
- do you have a single coordinate across multiple source files, or multiple coordinates all in a single source file
- how many orders-of-magnitude cutouts are you asking for: 10^2, 10^4, 10^6
- are you cutting out of L2 or L3 source files
- potentially: are you asking for FITS or ASDF output format

With such a matrix, then we should run analytics and analysis on the _current_ performance with profiling.  Such analysis should be scheduled earlier the better, but balancing with our other work.  The good news is we have plenty of L2 and L3 sample files to work with now.  Without doing such investigative work it's too early to know WHERE to look into further improvements, and I would not at all be surprised if the bottlenecks change between at least some of these permutations.

snbianco:
Yep, I see these changes as low-hanging fruit, and the profiling analysis will help us to identify bottlenecks and what, if anything, Astrocut can do to mitigate them. I did some rudimentary profiling while making these changes, and they shave off around a second when making 200x200 cutouts from 20 local L3 files. Not a huge gain, but it's there. The majority of execution time is coming from 1) opening the ASDF files and 2) decompressing the arrays.

braingram:
Thanks. Is it straightforward to share the scripts/test data you used for this profiling? I suspect there are ways to improve the load speed.

Since saving (and validation on save) was not the bottleneck would you restore the validation on save?

snbianco:
Sure! GitHub won't let me share the profiling files, but I can send them in a Slack message.

havok2063 (approved):
I think this work is good to merge.

PR #193, opened 2026-07-02
adds utils.utils.parse_table_wcs to resolve #191

### LINKED ISSUE #191: expose a public function for building a WCS from a cube's image-header table
[`CubeCutout._parse_table_info`](https://github.com/spacetelescope/astrocut/blob/18d5b4b/astrocut/cube_cutout.py#L116-L177) walks a cube's HDU-2 image-header table outward from the middle row, looking for the row whose WCS-axis column (e.g. `CTYPE2`) matches the expected value (e.g. `'DEC--TAN-SIP'` for TESS SPOC), then builds a `~astropy.wcs.WCS` object from that row. It's a private instance method that can only be reached by constructing a full `CubeCutout`/`TessCubeCutout` (which mandates `coordinates` and `cutout_size`).

The tesscut service has an endpoint (`/ffi_wcs`) that needs this logic, but with no target coordinates at all. It just wants the WCS for a given sector/camera/ccd cube, independent of any cutout. It is also used in the moving target cutouts endpoint.

Because `_parse_table_info` can't be reached without constructing a coordinate-bound cutout object, tesscut needs to duplicate the logic.

Can we expose a version of this row-search/WCS-building logic directly from astrocut, so tesscut doesn't need to duplicate? e.g.:

```py
def parse_table_wcs(table_data, wcs_axes_keyword, wcs_axes_value):
    """
    Find the row in a cube's image-header table (HDU 2 data) that contains a full WCS for the
    underlying FFI, and build a WCS object from it.
    """
```

Why this used to work

tesscut previously relied on a pattern like:

```py
astrocut_cutter = CutoutFactory()
astrocut_cutter._parse_table_info(table_data)
wcs = astrocut_cutter.cube_wcs
```

This worked because `CutoutFactory` was a TESS-specific class with a no-argument `__init__` and its own `_parse_table_info` method hardcoded `'CTYPE2'` / `'DEC--TAN-SIP'` inline and stored the result on `self.cube_wcs`. No coordinates were required to construct it or to call that method.

That stopped working in https://github.com/spacetelescope/astrocut/pull/146, which replaced that monolithic implementation with the current `CubeCutout`/ `TessCubeCutout` class hierarchy. `CutoutFactory` is now a wrapper around `TessCubeCutout` (its `__init__`/`_parse_table_info` were removed), and `_parse_table_info` moved to `CubeCutout`, whose `__init__` mandates `coordinates` and `cutout_size`.  As a result, `CutoutFactory()` no longer has a `_parse_table_info` method at all, and without this refactoring tesscut would have to initialize a TessCubeCutout instance, which performs the cutout immediately.

falkben:
adds utils.utils.parse_table_wcs to resolve #191


snbianco:
This looks good, but would it make more sense to make `parse_table_wcs` a static method in `CubeCutout` since the function is specific to TESS cube data? What are your thoughts?

falkben:
> This looks good, but would it make more sense to make `parse_table_wcs` a static method in `CubeCutout` since the function is specific to TESS cube data? What are your thoughts?

Agreed, but I moved it into `CubeCutout` as a static method instead, since it's relevant to CubeCutouts, not just TESS

This would let tesscut get the wcs still with this:

`wcs = TessCubeCutout.parse_table_wcs(table_data, "CTYPE2", "DEC--TAN-SIP")`

snbianco (approved):

PR #194, opened 2026-07-02
Add legacy_filenames parameter to cutout classes for pre-1.2.0 filename format. resolves #192

### LINKED ISSUE #192: option to restore the legacy TPF/cutout filename format
Cutout filenames changed format in https://github.com/spacetelescope/astrocut/pull/167. There's no way to opt back into the old format, and tesscut needs to reconstruct it to keep filenames stable for users.

| | Old (pre-1.2.0) | New (1.2.0+) |
|---|---|---|
| Format | `<stem>_<ra>_<dec>_<ny>x<nx>_astrocut.fits` | `<stem>_<ra>_<dec>_<ny>-x-<nx>_astrocut.fits` |
| RA/Dec precision | 6 decimals | 7 decimals |
| Example | `..._83.406310_-62.489771_64x64_astrocut.fits` | `..._83.4063100_-62.4897710_64-x-64_astrocut.fits` |

Could we add an opt-in parameter (e.g. `legacy_filenames: bool = False`) to the relevant classes/functions (`TessCubeCutout`, `TessFootprintCutout`, `CutoutFactory.cube_cut`, `cube_cut`) to restore the pre-1.2.0 format (`x` separator, no hyphen, 6-decimal RA/Dec)?

falkben:
resolves #192

### DIFF
--- a/astrocut/cube_cutout.py
+++ b/astrocut/cube_cutout.py
@@ -39,12 +39,12 @@ class CubeCutout(Cutout, ABC):
     threads : int | 'auto'
         The number of threads to use for making the cutouts. If 'auto', the number of threads will be set to the number
         of available CPUs.
+    verbose : bool
+        If True, log messages are printed to the console.
     legacy_filenames : bool
         If True, generated cutout filenames use the pre-1.2.0 format (``<ny>x<nx>`` size separator
         and 6-decimal RA/Dec precision) instead of the current format (``<ny>-x-<nx>`` size separator
         and 7-decimal RA/Dec precision). Default is False.
-    verbose : bool
-        If True, log messages are printed to the console.
 
     Attributes
     ----------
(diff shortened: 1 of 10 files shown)


astrocut/cube_cutout.py
```diff
@@ -65,8 +69,11 @@ def __init__(
         limit_rounding_method: str = "round",
         threads: Union[int, Literal["auto"]] = 1,
         verbose: bool = False,
+        legacy_filenames: bool = False,
```
snbianco: Rather than write `legacy_filenames` into the class constructor, could we make it a parameter on the write methods? I usually try to separate the parameters that impact the actual cutout process from those that are specific to written files.

falkben: yes, this makes total sense. done.

snbianco (approved):
Looks good to me!

PR #196, opened 2026-07-29
Accept multiple coordinates in ASDF Cutouts

snbianco:
This PR expands the ASDF cutout workflow to support multiple target coordinates per input file, instead of only a single coordinate. It also updates the cutout APIs to return structured tables of results, making it easier to inspect and work with cutouts across file/coordinate combinations.

- `ASDFCutout` can now accept a single coordinate or a list of coordinates. Cutouts are generated for each valid (input file, coordinate) pair.
- `cutouts`, `asdf_cutouts`, `fits_cutouts`, and `get_image_cutouts()` now expose results as `astropy.table.Table` objects for ASDF-based cutouts.
- Added iterator-based methods to lazily stream cutouts without materializing everything in memory.
- Added selection filters for input files and coordinates.
- `Cutout`, `FITSCutout`, `CubeCutout`, and TESS cutout classes were adjusted to align with the new coordinate handling. TESS footprint and cube cutout logic were updated accordingly.

This is a broader refactor of the ASDF cutout pipeline, so downstream code that previously expected lists of cutouts may need to be updated to work with the new table-based return values.

### DIFF
--- a/astrocut/asdf_cutout.py
+++ b/astrocut/asdf_cutout.py
@@ -955,7 +955,7 @@ class ASDFCutout(ImageCutout):
             asdf_file = nullcontext(file)
 
         with asdf_file as file_handle:
-            with asdf.open(file_handle) as af:
+            with asdf.open(file_handle, memmap=True) as af:
                 # Load the data from the input file
                 tree = af.tree
                 mission_tree = tree[self._mission_kwd] if self._mission_kwd in tree else None
--- a/astrocut/asdf_cutout.py
+++ b/astrocut/asdf_cutout.py
@@ -142,7 +142,8 @@ class ASDFCutout(ImageCutout):
     @property
     def asdf_cutouts(self) -> Table:
         """
-        Return the cutouts as a list of `asdf.AsdfFile` objects.
+        Return the cutouts as a `astropy.table.Table` with columns for input file, coordinate, and
+        the corresponding `asdf.AsdfFile` object.
         """
         if self._asdf_cutouts is not None:
             return self._asdf_cutouts
@@ -153,7 +154,8 @@ class ASDFCutout(ImageCutout):
     @property
     def fits_cutouts(self) -> Table:
         """
-        Return the cutouts as a list `astropy.io.fits.HDUList` objects.
+        Return the cutouts as a `astropy.table.Table` with columns for input file, coordinate, and
+        the corresponding `astropy.io.fits.HDUList` object.
         """
         if self._fits_cutouts is not None:
             return self._fits_cutouts
@@ -1043,6 +1045,7 @@ class ASDFCutout(ImageCutout):
                     # Make a per-coordinate copy because _get_cutout_data may modify the mission_tree
                     # in place if not in lite mode
                     coord_mission_tree = dict(new_mission_tree)
+                    coord_mission_tree["meta"] = dict(new_mission_tree["meta"])
 
                     try:
                         data_cutout = self._get_cutout_data(coord_mission_tree, wcs, pixel_coords)
@@ -1501,7 +1504,7 @@ def asdf_cut(
         )
 
 
-def get_center_pixel(gwcsobj: gwcs.wcs.WCS, ra: float, dec: float) -> Tuple[Tuple[int, int], WCS]:
+def get_center_pixel(gwcsobj: gwcs.wcs.WCS, ra: float, dec: float) -> Tuple[int, int]:
     """
     Get the closest pixel location on an input image for a given set of coordinates.
 
@@ -1517,7 +1520,7 @@ def get_center_pixel(gwcsobj: gwcs.wcs.WCS, ra: float, dec: float) -> Tuple[Tupl
     Returns
     -------
     pixel_coords : tuple
-        The (row, col) pixel coordinates of the input coordinates.
+        The (row, col) pixel coordinates of the input coordinates on the image.
     """
     # Map the coordinates to a pixel's location on the 2d image
     row, col = gwcsobj.invert(np.atleast_1d(ra), np.atleast_1d(dec), with_bounding_box=False)
--- a/astrocut/asdf_cutout.py
+++ b/astrocut/asdf_cutout.py
@@ -1,11 +1,11 @@
 import sys
 import warnings
 from contextlib import nullcontext
-from copy import copy
+from copy import deepcopy
 from datetime import date
 from pathlib import Path
 from time import monotonic
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import asdf
 import gwcs
@@ -15,14 +15,16 @@ from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.modeling import models
 from astropy.nddata.utils import Cutout2D, NoOverlapError
+from astropy.table import Column, Table
 from astropy.units import Quantity
 from astropy.utils.decorators import deprecated_renamed_argument
 from astropy.wcs import WCS
 from packaging.version import Version
+from PIL.Image import Image, Transpose, fromarray
 from s3path import S3Path
 
 from . import __version__, log
-from .exceptions import DataWarning, InvalidInputError, InvalidQueryError, ModuleWarning
+from .exceptions import DataWarning, InputWarning, InvalidInputError, InvalidQueryError, ModuleWarning
 from .image_cutout import ImageCutout
 
 
@@ -34,8 +36,8 @@ class ASDFCutout(ImageCutout):
     ----------
     input_files : list
         List of input image files.
-    coordinates : str | `~astropy.coordinates.SkyCoord`
-        Coordinates of the center of the cutout.
+    coordinates : str | `~astropy.coordinates.SkyCoord` | list
+        Coordinates of the center of the cutout. You can pass a single coordinate or a list of coordinates.
     cutout_size : int | array | list | tuple | `~astropy.units.Quantity`
         Size of the cutout array.
     fill_value : int | float
@@ -83,7 +85,7 @@ class ASDFCutout(ImageCutout):
     def __init__(
         self,
         input_files: List[Union[str, Path, S3Path]],
-        coordinates: Union[SkyCoord, str],
+        coordinates: Union[SkyCoord, str, List[Union[SkyCoord, str]]],
         cutout_size: Union[int, np.ndarray, Quantity, List[int], Tuple[int]] = 25,
         fill_value: Union[int, float] = np.nan,
         key: Optional[str] = None,
@@ -92,7 +94,6 @@ class ASDFCutout(ImageCutout):
         lite: Optional[bool] = True,
         verbose: bool = False,
     ):
-        # Superclass constructor
         super().__init__(input_files, coordinates, cutout_size, fill_value, verbose=verbose)
 
         # Must be using Python 3.11 or higher to support stdatamodels and ASDF-in-FITS embedding
@@ -108,47 +109,13 @@ class ASDFCutout(ImageCutout):
         self.cutouts = []  # Public attribute to hold `Cutout2D` objects
         self._asdf_cutouts = None  # Store ASDF objects
         self._fits_cutouts = None  # Store FITS objects
-        self._sliced_gwcs_objects = []  # Store sliced GWCS objects for lite mode
-        self._asdf_trees = []  # Store ASDF trees for each cutout
+        self._asdf_trees = {}  # Store ASDF trees for each cutout
         self._lite = lite  # Flag for lite mode
         self._primary_header_template = None  # Optional template for primary header keywords
         self._fill_value_cache = {}  # Cache for converted fill values based on input data types
 
         # Make cutouts
         self.cutout()
-
-    def _check_asdf_in_fits_support(self):
-        if self._asdf_in_fits is not None:
-            return
-
-        # Try to import stdatamodels for ASDF-in-FITS embedding
-        if self._py311_or_higher:
-            try:
-                # Check version of stdatamodels
-                from stdatamodels import __version__ as stdata_version
-                from stdatamodels import asdf_in_fits
-
-                if Version(stdata_version) < Version("4.1.0"):
-                    warnings.warn(
-                        "The `stdatamodels` package is not available in the correct version (>=4.1.0); "
-                        "ASDF-in-FITS embedding will be skipped for these cutouts. Install the optional "
-                        'dependency with: pip install "astrocut[all]" or pip install stdatamodels>=4.1.0',
-                        ModuleWarning,
-                    )
-                else:
-                    self._asdf_in_fits = asdf_in_fits
-            except ImportError:
-                warnings.warn(
-                    "The `stdatamodels` package cannot be imported; ASDF-in-FITS embedding will be "
-                    "skipped for these cutouts. Install the optional dependency with: "
-                    'pip install "astrocut[all]" or pip install stdatamodels>=4.1.0',
-                    ModuleWarning,
-                )
-        else:
-            warnings.warn(
-                "ASDF-in-FITS embedding requires Python 3.11 or higher. Skipping embedding for these cutouts.",
-                ModuleWarning,
-            )
 
     @property
     def fits_cutouts(self) -> List[fits.HDUList]:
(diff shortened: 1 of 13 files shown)


docs/astrocut/index.rst
```diff
@@ -98,7 +98,8 @@ The Advanced Scientific Data Format (ASDF) is a flexible format for storing scie
 and return the results in memory or as a written file, depending on the user's preference. The cutout ASDF file format is
 described in the :ref:`Astrocut File Formats <asdf-cutout-files>` page.
 
-To make a cutout from an ASDF file or files, use the `~astrocut.ASDFCutout` class.
+To make cutouts from an ASDF file or files, use the `~astrocut.ASDFCutout` class. The ``coordinates`` argument can be a
+single coordinate or a list of coordinates, and Astrocut will generate one cutout for each coordinate in each input file.
```
havok2063, line 102: is the input strictly a single or set of `SkyCoord` objects, or can the user input a list of string coordinates?

docs/astrocut/index.rst
```diff
@@ -107,40 +108,64 @@ To make a cutout from an ASDF file or files, use the `~astrocut.ASDFCutout` clas
 
   >>> input_files = [""]  # Path(s) to local ASDF file, URL, or S3 URI
 
-  >>> center_coord = SkyCoord("80.15189743 29.74561219", unit="deg")
+  >>> coords = [SkyCoord("80.15189743 29.74561219", unit="deg"),
+  ...           SkyCoord("80.15500000 29.75000000", unit="deg")]
```
havok2063, line 112: For large sets of coordinates that get operated on, astropy recommends a single SkyCoord with an array of coords, e.g. 
```
coords=['80.15189743 29.74561219', '80.15500000 29.75000000']
skyc = SkyCoord(coords, unit='degree')
<SkyCoord (ICRS): (ra, dec) in deg
    [(80.15189743, 29.74561219), (80.155     , 29.75      )]>
```
If a user is operating on all sources from a Roman catalog, this may be what they input.  If we don't support this already, we may want to.

docs/astrocut/index.rst
```diff
@@ -107,40 +108,64 @@ To make a cutout from an ASDF file or files, use the `~astrocut.ASDFCutout` clas
...
   cutouts that are more accurately centered on the target coordinates than even values
   for ``cutout_size``.
 
-The resulting `~astrocut.ASDFCutout` object can be used to access the cutout science data and metadata. The ``cutouts_by_file`` attribute is a dictionary that
-stores the individual science data cutouts as a list of `~astropy.nddata.Cutout2D` objects by input filename. The `~astropy.nddata.Cutout2D`
-object contains the science cutout data, shape, world coordinate system (WCS) and other helpful properties. The ``cutouts`` attribute is a list of
-science data cutouts as `~astropy.nddata.Cutout2D` objects, one for each input file.
+The resulting `~astrocut.ASDFCutout` object can be used to access the cutout science data and metadata.
+The ``cutouts_by_file`` attribute is a nested dictionary keyed by input filename and coordinate string,
+where each value is a `~astropy.nddata.Cutout2D` object.
```
havok2063, line 125: Can you add an example of the output of this attribute?

docs/astrocut/index.rst
```diff
@@ -107,40 +108,64 @@ To make a cutout from an ASDF file or files, use the `~astrocut.ASDFCutout` clas
...
 
-The ``asdf_cutouts`` attribute is a list of cutouts as `~asdf.AsdfFile` objects, and the ``fits_cutout`` attribute is a list of cutouts as
-`~astropy.io.fits.HDUList` objects. The cutout objects in these lists can be used to access cutout data and metadata, as shown below.
+The ``cutouts`` attribute is an `~astropy.table.Table` with columns ``file``, ``coordinate``, and ``cutout``.
+The ``cutout`` column stores `~astropy.nddata.Cutout2D` objects, with one row per valid
+``(input file, coordinate)`` pair.
```
havok2063: We might want to clarify here the table contains one row per unique cutout, defined by a (input file, coordinate) pair.

astrocut/asdf_cutout.py
```diff
@@ -105,19 +106,270 @@ def __init__(
...
+        Returns
+        -------
+        cutouts : `astropy.table.Table`
+            Table with columns for input file, coordinate, and the corresponding `astropy.nddata.Cutout2D` object.
+        """
+        cutout_table = Table(
+            rows=list(self.iter_cutouts(input_files=input_files, coordinates=coordinates)),
+            names=["file", "coordinate", "cutout"],
+        )
+        return cutout_table
```
havok2063: We can return the Table directly

astrocut/asdf_cutout.py
```diff
@@ -105,19 +106,270 @@ def __init__(
...
+        -------
+        asdf_cutouts : `astropy.table.Table`
+            Table with columns for input file, coordinate, and the corresponding `asdf.AsdfFile` object representing
+            the cutout.
+        """
+        asdf_cutout_table = Table(
+            rows=list(self.iter_asdf_cutouts(input_files=input_files, coordinates=coordinates)),
+            names=["file", "coordinate", "cutout"],
+        )
+        return asdf_cutout_table
```
havok2063:
```suggestion
        return Table(
            rows=list(self.iter_asdf_cutouts(input_files=input_files, coordinates=coordinates)),
            names=["file", "coordinate", "cutout"],
        )
```

astrocut/asdf_cutout.py
```diff
@@ -105,19 +106,270 @@ def __init__(
...
+        coordinates : list
+            Optional. List of coordinates to include in the output. If not specified, all coordinates will be included.
+
+        Returns
+        -------
+        cutouts : `astropy.table.Table`
+            Table with columns for input file, coordinate, and the corresponding `astropy.nddata.Cutout2D` object.
+        """
+        cutout_table = Table(
+            rows=list(self.iter_cutouts(input_files=input_files, coordinates=coordinates)),
```
havok2063: You don't need to convert the generator to a list first.  `rows`  will accept it directly.

astrocut/asdf_cutout.py
```diff
@@ -105,19 +106,270 @@ def __init__(
...
+            Optional. List of coordinates to include in the output. If not specified, all coordinates will be included.
+
+        Returns
+        -------
+        asdf_cutouts : `astropy.table.Table`
+            Table with columns for input file, coordinate, and the corresponding `asdf.AsdfFile` object representing
+            the cutout.
+        """
+        asdf_cutout_table = Table(
+            rows=list(self.iter_asdf_cutouts(input_files=input_files, coordinates=coordinates)),
```
havok2063: same here

astrocut/asdf_cutout.py
```diff
@@ -105,19 +106,270 @@ def __init__(
...
+        """
+        fits_cutouts = list(self.iter_fits_cutouts(input_files=input_files, coordinates=coordinates))
+
+        fits_cutout_table = Table()
+        fits_cutout_table["file"] = [item[0] for item in fits_cutouts]
+        fits_cutout_table["coordinate"] = [item[1] for item in fits_cutouts]
+        fits_cutout_table.add_column(Column(length=len(fits_cutouts), dtype=object, name="cutout"))
+        for i, item in enumerate(fits_cutouts):
+            fits_cutout_table["cutout"][i] = item[2]
+        return fits_cutout_table
```
havok2063: Can this be simplified to the same syntax as in `get_cutouts` and `get_asdf_cutouts`?

astrocut/asdf_spectral_subset.py
```diff
@@ -135,7 +135,7 @@ def _subset_single_file(
     out_trees_for_file = {}
     emitted_warnings = []
 
-    with asdf.open(file) as af:
+    with asdf.open(file, memmap=True) as af:
```
havok2063, line 138: Should this, and other asdf kwargs, be configurable by the user, with defaults set?

astrocut/image_cutout.py
```diff
@@ -191,63 +192,160 @@ def get_image_cutouts(
...
+        """
+        Build a standard image cutout output table from iterator rows.
+        """
+        image_table = Table()
+        image_table["file"] = [row[0] for row in image_rows]
+        image_table["coordinate"] = [row[1] for row in image_rows]
+        image_table.add_column(Column(length=len(image_rows), dtype=object, name="cutout"))
+        for i, row in enumerate(image_rows):
+            image_table["cutout"][i] = row[2]
+        return image_table
```
havok2063: similar question as with the fits table, is there a reason why this syntax differs from the asdf/cutout table creation?

astrocut/asdf_cutout.py
```diff
@@ -150,97 +402,285 @@ def _check_asdf_in_fits_support(self):
...
-                    "homepage": "https://astrocut.readthedocs.io/en/latest/",
-                },
-            )
-            asdf_cutouts.append(af)
+        file_coord_pairs = []
+        for file in files_to_include:
+            for coord in coords_to_include:
+                if coord not in self.cutouts_by_file[file]:
+                    continue  # Skip coordinates that are not associated with this file
+                file_coord_pairs.append((file, coord))
```
havok2063: This method could probably also just be an iterator.

astrocut/asdf_cutout.py
```diff
@@ -150,97 +402,285 @@ def _check_asdf_in_fits_support(self):
...
+                else:
+                    raise InvalidInputError(f"Coordinate {coord} is not a valid SkyCoord or string.")
+
+            for coord in coords_to_include:
+                if coord not in all_coords:
+                    raise InvalidInputError(f"Input coordinate {coord} is not in the cutout results.")
+
+        return files_to_include, coords_to_include
+
+    def cutout(self) -> Union[str, List[str], List[fits.HDUList]]:
```
havok2063, line 701: Should this be called something like `generate_cutouts` since it produces cutouts?  To differentiate from the `cutouts` which are the set of cutout objects.

docs/astrocut/index.rst
```diff
@@ -107,40 +108,64 @@ To make a cutout from an ASDF file or files, use the `~astrocut.ASDFCutout` clas
...
+  ...     input_files=input_files[:1],
+  ...     coordinates=coords[:1],
+  ... )
+  >>> len(subset) #doctest: +SKIP
+  1
+  >>> subset["cutout"][0].data.shape #doctest: +SKIP
+  (25, 25)
+
+The ``asdf_cutouts`` and ``fits_cutouts`` attributes also both return `~astropy.table.Table` objects
+with columns ``file``, ``coordinate``, and ``cutout``. The ``cutout`` column stores `~asdf.AsdfFile`
```
havok2063, line 164: This mean that the output `ASDFCutout` has many attributes that return tables, that only differ by the contents of `cutout`?   Should this instead be a single output table with 4 columns, with one for the cutout and one for the requested format output?

docs/astrocut/index.rst
```diff
@@ -181,21 +206,43 @@ cutout FITS files.
...
+
+Streaming Cutouts
+^^^^^^^^^^^^^^^^^^
+
+For large batches, you can stream cutouts lazily using iterator methods rather than materializing full tables in memory:
+
+- ``iter_cutouts(...)`` yields tuples of ``(file, coordinate, cutout2D)``.
+- ``iter_asdf_cutouts(...)`` yields tuples of ``(file, coordinate, asdf_cutout)``.
+- ``iter_fits_cutouts(...)`` yields tuples of ``(file, coordinate, fits_cutout)``.
+- ``iter_image_cutouts(...)`` yields tuples of ``(file, coordinate, image_cutout)``.
```
havok2063, line 231: It looks like these just iterate over the cutouts to format the output, correct?  And not iteratively produce the cutouts themselves?  Do we have a sense of the overhead of producing all the cutouts at once first, then produce outputs vs performing the cutout only during outputting the format?

havok2063 (commented):
There are at least 4 different Table calls.  AFAIK, the only difference is the cutout column containing the object.  If we ever need to make changes to the output table, that's a lot of changes in places.  Is it possible to consolidate those down into a single code handler for Table output?  Then each format fxn only updates the one column?

havok2063:
Overall it looks good.  I really like the addition of the iterators.  It makes me wonder if we should try to iterate over as much as possible.

docs/astrocut/index.rst
snbianco: Huh, I had not considered this 🤔 My only concerns would be 1) processing time to create all of the objects that would be in that one giant table and 2) it wouldn't be compatible with the existing API (but I'm less concerned about this for `ASDFCutout`). I do like this idea to make things simpler though. What do you think about a potential performance hit and is there a way we could work around that?

docs/astrocut/index.rst
snbianco: Do you mean iteratively produce the cutouts when a user requests them in a certain format? I'm not sure that would work with how Astrocut is currently set up. For example, if a user were to request a cutout from a really large image that takes 10 s to open from a specific coordinate, and then later request a different location from the same image, I think that would mean the image would have to be opened twice (or kept open in memory). Producing all the cutouts first lets us make the most of each open file handler and be able to close them once done.

astrocut/asdf_cutout.py
snbianco: I agree that it's confusing, but this is an inherited method coming from the base class. If it helps, the user should never really have to call the `cutout` method, since it's called in the constructor.

astrocut/image_cutout.py
snbianco: There was some trickiness where if I used the same process with FITS/images as I did with ASDF, the data type for the column was not specific and just "object". After thinking about it some more, I'm not sure how much this really matters.

docs/astrocut/index.rst
snbianco: Good point, I'll add support for that.

docs/astrocut/index.rst
snbianco: Yes, they can input a list of string coordinates as well and Astrocut will convert them to `SkyCoord`.

snbianco:
I've also realized that for the `get_{}_cutouts` methods, we could totally use the cached tables if they already exist, and then just filter by input file and coordinate. I'll make that change in the next commit as well.

astrocut/asdf_cutout.py
snbianco: There's an issue where if we use the same syntax, the cutout column sets its datatype to an array rather than an `HDUList`. I'll add a clarifying comment to the code.

docs/astrocut/index.rst
havok2063: Yeah I do worry about the performance hit, but what is the performance impact now of creating 4 tables?  I suppose they are on demand.  I guess I'm wondering if we could consolidate the Table logic into a singular place, for easier updating in the future?  Or cache an initial table of file, coordinate, (optionally cutout); then when a user calls `get_xxx_cutout` it only does the processing to add the cutout column or the 4th object column?  

The reason I split `cutout` and `object` is in case the user wants the cutout object in addition to the file cutout object.

astrocut/image_cutout.py
havok2063: Is this `image_table.add_column(Column(length=len(image_rows), dtype=object, name="cutout")` setting the dtype to non-specific "object" as well?   I don't fully understand how that is different.  I'm not sure how much the Table and Column have changed in recent astropy versions, but dtype might only be used for units and type conversions.

astrocut/asdf_cutout.py
havok2063: Ahh I see.  I do see that could be confusing for users.  But if they inspect a row value in the cutout, would it still be an HDUList object?

docs/astrocut/index.rst
havok2063: Yeah that's what I meant, but I forgot about the performance hit of opening the file.  The only option would be to cache all open files until the iterator completes, which I'm not sure is something we want to do automatically.

astrocut/asdf_cutout.py
snbianco: No, the type of the value in the cells is a numpy array with the simplified syntax.

astrocut/image_cutout.py
snbianco: The issue with the image objects when using the simplified table construction is actually an issue in the astropy `Column` code. It looks like it errors out because Astropy tries to iterate over the `Image` object.

snbianco:
Ok, looks like CI is working now.

havok2063:
@snbianco Are you still working on this?  Do you want me to do another pass or look at anything in particular?

snbianco:
Sorry, don't know how I'm just now seeing this! Yes, just needs another round of review for the changes.

havok2063 (approved):
I think this is good to go.  

One thing that came out of the profiling was that while the astropy Table creation is generally negligible, and appending a row adds very minimal time and memory, producing a large Table of millions of rows in memory could add significant overhead.  We may want to strategize a bit on how this should work on the RRN, with limits or batches, or what have you. I think we can do this, and any other follow-up work, in future PRs.  No need to hold this one up.

